PR #546, opened 2014-12-31
Respect original property ordering.

trecloux:
This pull request is following PR #543 from my colleague @gwallet .
We finally found a way to respect the original property ordering.

This pull request includes a unit test.

Thanks, Thomas.

### DIFF
--- a/swagger-models/src/test/groovy/com/mangofactory/swagger/models/property/bean/BeanModelPropertyProviderSpec.groovy
+++ b/swagger-models/src/test/groovy/com/mangofactory/swagger/models/property/bean/BeanModelPropertyProviderSpec.groovy
@@ -1,0 +1,36 @@
+package com.mangofactory.swagger.models.property.bean
+
+import com.fasterxml.classmate.ResolvedType
+import com.fasterxml.classmate.TypeResolver
+import com.fasterxml.jackson.databind.ObjectMapper
+import com.mangofactory.swagger.mixins.ModelPropertyLookupSupport
+import com.mangofactory.swagger.mixins.TypesForTestingSupport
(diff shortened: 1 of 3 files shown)


dilipkrish:
@trecloux thanks for the fix! Still trying to understand how filtering the method with a check on the property name manages to sort it correctly. :confused: 

Before I pull it in, I just want to make sure we don't have a gap in the existing test cases that it may regress.

swagger-models/src/test/groovy/com/mangofactory/swagger/models/property/bean/BeanModelPropertyProviderSpec.groovy
```diff
@@ -0,0 +1,36 @@
...
+    given:
+      Class typeToTest = complexType()
+      def typeResolver = new TypeResolver()
+      ResolvedType resolvedType = typeResolver.resolve(typeToTest)
+      ObjectMapper mapper = new ObjectMapper()
+
+      def beanModelPropertyProvider = new BeanModelPropertyProvider(new AccessorsProvider(typeResolver), typeResolver,
+            new AlternateTypeProvider(), new ObjectMapperBeanPropertyNamingStrategy(mapper))
+      beanModelPropertyProvider.objectMapper = mapper
+      def properties = beanModelPropertyProvider.propertiesForSerialization(resolvedType)
```
dilipkrish: speficifically, does the following also work, for weird edge cases in the way jackson allows you to configure beans! :)

```java
def properties = beanModelPropertyProvider.propertiesForDeserialization(resolvedType)
```

dilipkrish:
Thanks @trecloux! :+1:

Issue #552, opened 2015-01-07
ApiResponses annotation is ignored when placed on an interface


jlamby:
To keep implementation code as short as possible, I try to put all annotations on the rest controller interface. It works with `@Api` and `@ApiOperation` but not with `@ApiResponses`. 
Am I missing something (like enabling something in SpringSwaggerConfig) ?

**Sample of not working ApiResponses :**

```java
@RequestMapping("service")
@Api(value="service", description="Operations on service")
public interface SimpleService {

    @RequestMapping(method = RequestMethod.POST, consumes = MediaType.APPLICATION_JSON_VALUE, produces = MediaType.APPLICATION_JSON_VALUE)
    @ApiOperation(value="Register a new client", response=RegistrationDTO.class)
    @ApiResponses({
        @ApiResponse(code=201, message="Client successfuly registered"),
        @ApiResponse(code=400, message="Bad request")
    })
    ResponseEntity<RegistrationDTO> register(InformationDTO information);
}

@RestController
public class SimpleServiceImpl implements SimpleService {

    @Override
    public ResponseEntity<RegistrationDTO> register(@Validated @RequestBody InformationDTO information) {
        // (...)
    }
}
```
*Swagger output :*
```javascript
"responseMessages": [
    { "code": 200, "message": null, "responseModel": "RegistrationDTO" },
    { "code": 201, "message": "Created" },
    { "code": 401, "message": "Unauthorized" },
    { "code": 403, "message": "Forbidden" },
    { "code": 404, "message": "Not Found" }
],
```

**Sample of working ApiResponses :**

```java
@RequestMapping("service")
@Api(value="service", description="Operations on service")
public interface SimpleService {

    @RequestMapping(method = RequestMethod.POST, consumes = MediaType.APPLICATION_JSON_VALUE, produces = MediaType.APPLICATION_JSON_VALUE)
    @ApiOperation(value="Register a new client", response=RegistrationDTO.class)
    ResponseEntity<RegistrationDTO> register(InformationDTO information);
}

@RestController
public class SimpleServiceImpl implements SimpleService {

    @Override
    @ApiResponses({
        @ApiResponse(code=201, message="Client successfuly registered"),
        @ApiResponse(code=400, message="Bad request")
    })
    public ResponseEntity<RegistrationDTO> register(@Validated @RequestBody InformationDTO information) {
        // (...)
    }
}
```
*Swagger output :*
```
"responseMessages": [
[...]

PR #555, opened 2015-01-08
[ #552] Enable ApiResponses annotation on interface

jlamby:


dilipkrish:
@jlamby Thanks for the PR!!

Closes #552

PR #559, opened 2015-01-09
Enum parser retains whitespace and brackets

ssaarela:
Having allowableValues="[Polygon, Point]" results in enum: ["[Polygon", " Point]"]. I was expecting ["Polygon", "Point"].

Probably the simplest - but not necessarily 100% correct - way to fix this is to handle also space, [ and ] as delimiters to split on.

(Sorry about the hasty pull request. Should I try again according to the guidelines?)


dilipkrish:
@ssaarela absolutely.. let me know if you need help with rebasing that commit.

dilipkrish:
@ssaarela :wave:

ssaarela:
We were able to work around this bug and it seems that I have no time for proper contribution. We also need Swagger 2.0 (because of GeoJSON), so this fix wouldn't help much any way - for now at least.

dilipkrish:
Closing this as its been fixed

PR #561, opened 2015-01-12
Use @ApiModel.value as alternate type name for serialization

### LINKED ISSUE #382: @ApiModel.value is not being honored as an alternate type name
The swagger core annotation `ApiModel` allows you to define [an alternate name](https://github.com/wordnik/swagger-core/blob/master/modules/swagger-annotations/src/main/java/com/wordnik/swagger/annotations/ApiModel.java#L34) for the type via the `value` field. This is not being honored. The description is being applied, but the type name is only inferred from the class name.

I think the Model id and name should prefer the `@ApiModel.value` in `DefaultModelProvider`, much like we pull the model description from `@ApiModel.description`.

handavid:
Replaces the ResolvedType with @ApiModel.value if present

closes #382


dilipkrish:
:+1: Thanks @handavid

Issue #588, opened 2015-02-12
Support @RequestPart annotation


ammmze:
TLDR: `@RequestPart` parameters should have `paramType` of `form` and models should be generated for `@RequestPart` annotated parameters.

---

When using the `@RequestPart` annotation, the paramType attribute for that parameter is incorrect (unless the parameter is a `MultipartFile`). When making an api operation that uses a file upload, we can no longer use `@RequestBody`, instead it must be `@RequestPart`, which means the overall request `Content-Type` becomes `mulitpart/form-data`.

While this all works fine if the if the file upload is the only parameter, once you try to add additional parameters (i.e. pieces that will be included in its own boundary), it no longer works (at least not with swagger-ui, because the spec is incorrect).

The other issue is the parameters annotated as `@RequestPart` do not have their models added to the generated spec files.

So given the following example, swagger-ui should create a form that takes a text input and file input (which it does), however the request parameter is using paramType 'body', and when the form is submitted, only the file gets submitted in the request.

```java
@ApiOperation(value = "")
@RequestMapping(value = "/issue", method = RequestMethod.POST)
@ResponseBody
public MyResponse issueRequest(
        @ApiParam(required = true) @RequestPart MyRequest request,
        @ApiParam(required = true) @RequestPart MultipartFile document)
```

Using the custom annotation readers, I am able to work around the paramType issue, but not adding the model to the spec issue.

zerkowsm:
It seems Swagger does NOT display Model | Example value section for @RequestPart in 2.7.0 version. It just displays undefined as data Type next to the text field. It works if you change @RequestPart into @RequestBody.

dilipkrish:
@zerkowsm this is fixed via #1965 in the next release

zerkowsm:
Thanks for the information @dilipkrish . When do you plan to release that?

dilipkrish:
Hopefully in a week. trying to round out some bugs before that.

zerkowsm:
:+1:

ndorigatti:
Hello, i see again this issue in springfox-boot-starter:3.0.0

I have an endpoint with both Pojo and Multipart file.
I can successfully call the endpoint with Postman tool if I set it:
` @PostMapping(value = "/sito", consumes = {MediaType.MULTIPART_FORM_DATA_VALUE})
    public void insertSito(@RequestPart(name = "edifici", required = false) MultipartFile file,
[...]

PR #589, opened 2015-02-12
Add support for @RequestPart annotation 

ammmze:
Add support for `@RequestPart` annotation which means the request content-type would be multipart/form-data, which must have paramType of 'form' in swagger-ui.

This is for Issue #588

Some caveats I've found when using `@RequestPart` on things other than files...

* Currently swagger-ui renders paramType 'form' as `<input type="text" />` instead of `<textarea>`. I plan to submit a pull request to fix that as well.
* Through swagger-ui, the content-type for the `@RequestPart`'s that aren't files are not set (limitation of HTML forms), so in order for spring to properly handle a request with a `@RequestPart` that isn't a file, I had to use a custom MultipartResolver to detect if the content type should be `application/json` if it was null.

### DIFF
--- a/swagger-springmvc/src/main/java/com/mangofactory/swagger/readers/ApiModelReader.java
+++ b/swagger-springmvc/src/main/java/com/mangofactory/swagger/readers/ApiModelReader.java
@@ -23,6 +23,7 @@ import org.slf4j.LoggerFactory;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.stereotype.Component;
 import org.springframework.web.bind.annotation.RequestBody;
+import org.springframework.web.bind.annotation.RequestPart;
 import org.springframework.web.method.HandlerMethod;
 
 import java.lang.annotation.Annotation;
@@ -182,7 +183,7 @@ public class ApiModelReader implements Command<RequestMappingContext> {
     for (int i = 0; i < annotations.length; i++) {
       Annotation[] pAnnotations = annotations[i];
       for (Annotation annotation : pAnnotations) {
-        if (annotation instanceof RequestBody) {
+        if (annotation instanceof RequestBody || annotation instanceof RequestPart) {
           ResolvedMethodParameter pType = parameterTypes.get(i);
           if (!settings.getIgnorableParameterTypes()
                   .contains(pType.getResolvedParameterType().getErasedType())) {
--- a/swagger-springmvc/src/main/java/com/mangofactory/swagger/readers/operation/parameter/ParameterTypeReader.java
+++ b/swagger-springmvc/src/main/java/com/mangofactory/swagger/readers/operation/parameter/ParameterTypeReader.java
@@ -11,6 +11,7 @@ import org.springframework.web.bind.annotation.PathVariable;
 import org.springframework.web.bind.annotation.RequestBody;
 import org.springframework.web.bind.annotation.RequestHeader;
 import org.springframework.web.bind.annotation.RequestParam;
+import org.springframework.web.bind.annotation.RequestPart;
 import org.springframework.web.multipart.MultipartFile;
 
 import java.lang.annotation.Annotation;
@@ -45,6 +46,8 @@ public class ParameterTypeReader implements Command<RequestMappingContext> {
           return "query";
         } else if (annotation instanceof RequestHeader) {
           return "header";
+        } else if (annotation instanceof RequestPart) {
+          return "form";
         }
       }
     }
--- a/swagger-springmvc/src/test/groovy/com/mangofactory/swagger/readers/ApiModelReaderSpec.groovy
+++ b/swagger-springmvc/src/test/groovy/com/mangofactory/swagger/readers/ApiModelReaderSpec.groovy
@@ -217,6 +217,63 @@ class ApiModelReaderSpec extends Specification {
 
   }
 
+  def "should only generate models for request parameters that are annotated with Springs RequestPart"() {
+    given:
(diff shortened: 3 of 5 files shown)


dilipkrish:
@ammmze :+1: I thought ```@RequestPart``` automatically hands it of to a http message converter to figure out the content type. Are you saying its a limitation in the swagger-ui to not be able to handle variable content types in a multipart-form? 

I'm wondering if it should have a parameter type of _form_ instead?

ammmze:
@dilipkrish, It would hand it off to the message converter if the Content-Type for the boundary was set, however a Content-Type is not set at all (unless its a file).

Example request:
```
{other headers}
Content-Type:multipart/form-data; boundary=----WebKitFormBoundarydGA65oLFZj9mGBg5
{other headers}

------WebKitFormBoundarydGA65oLFZj9mGBg5
Content-Disposition: form-data; name="parameter1"

{}
------WebKitFormBoundarydGA65oLFZj9mGBg5
Content-Disposition: form-data; name="parameter2"; filename="example logo 800x600_382_blogImage_jpg_l.jpg"
Content-Type: image/jpeg


------WebKitFormBoundarydGA65oLFZj9mGBg5--
```

Currently in swagger-ui I do not see any way that you can specify the Content-Type for a specific parameter/boundary. And yes, the parameter type must be 'form' (or in swagger 2.0 it should actually be 'formData' apparently). Setting the parameter type to 'form' is one of the goals of this pull request. Without this change, the generated spec produces `parameter1` as `body` and `parameter2` as `form` and when the request is sent, only `parameter2` is in the payload.

ammmze:
Back on the message converter issue, just before the parameter is passed to the message converter, it goes to `org.springframework.web.multipart.support.RequestPartServletServerHttpRequest` which ends up throwing an exception if the parameter has not headers.

If you (or anybody) is curious, this is what I came up with for overcoming the lack of Content-Type header in the boundary. Note: I currently only support JSON, so it only detects JSON.

```java
public class DetectContentTypeMultipartResolver extends CommonsMultipartResolver {

    @Override
    protected MultipartParsingResult parseRequest(HttpServletRequest request) throws MultipartException {
        MultipartParsingResult result = super.parseRequest(request);

        // Detects the content type of the multipart/form-data if the content type is not specified
        Map<String, String> contentTypes = result.getMultipartParameterContentTypes();
        for (Map.Entry<String, String> entry : contentTypes.entrySet()) {
            if (entry.getValue() == null) {
                String[] values = result.getMultipartParameters().get(entry.getKey());
                for (String value : values) {
                    if (value != null) {
                        value = value.trim();
                        if ((value.startsWith("{") && value.endsWith("}"))
                                || (value.startsWith("[") && value.endsWith("]"))) {
                            contentTypes.put(entry.getKey(), "application/json");
                            break;
                        }
                    }
                }
            }
        }
        return result;
    }
}
```

dilipkrish:
@ammmze Merging this in.. would you be able to shepherd this in the 2.0 version aswell? [see WIP here](https://github.com/martypitt/swagger-springmvc/tree/refactor/538/spike-to-seperate-models-and-dtos)

ammmze:
I haven't played in the 2.0 code, mind pointing me in the right direction? I'm seeing readers in a few different packages.

dilipkrish:
@ammmze Thanks for taking a look! :+1: 

For starters look at the [OperationModelsProvider](https://github.com/martypitt/swagger-springmvc/blob/refactor/538/spike-to-seperate-models-and-dtos/mangodocs-spring-web/src/main/java/com/mangofactory/documentation/spring/web/readers/operation/OperationModelsProvider.java#L64) and also [ParameterTypeReader](https://github.com/martypitt/swagger-springmvc/blob/refactor/538/spike-to-seperate-models-and-dtos/mangodocs-spring-web/src/main/java/com/mangofactory/documentation/spring/web/readers/parameter/ParameterTypeReader.java#L42)

Just a little bit of background. We've separated the spring specific processing and the swagger specific processing. Earlier it was all bundled into one of two modules (swagger-springmvc or swagger-models). Now we've layered how the service model is inferred and the whole service model comes together as a result of layering a set of plugins.

So the flow is now like this
```ascii
                          +--------------------------+                             
+---------------------+   |  Plugins that infer      |                             
|                     |   |  service models based    |                             
| spring mvc service  +--->  on spring mvc request   +---------------+             
|                     |   |  mappings and annotations|               |             
+---------------------+   |                          |               |             
                          +------------+-------------+               |             
                                       |                             |             
                                       |                             |             
                                       |                             |             
                          +------------v------------+   +------------+------------+
                          |  Plugins that augment   |   |  Plugins that augment   |
                          |  service models based   |   |  service models based   |
                          |  on swagger v1.2        |   |  on swagger v2.0        |
                          |  annotations            |   |  annotations            |
                          +------------+------------+   +------------+------------+
                                       |                             |             
                                       |                             |             
                          +------------v-----+------+   +------------v-----+------+
                          |  Mappers that convert   |   |  Mappers that convert   |
                          |  inferred models to     |   |  inferred models to     |
                          |  swagger DTOs per v1.2  |   |  swagger DTOs per v2.0  |
                          |  of the spec            |   |  of the spec            |
                          +------------+------------+   +------------+------------+
                                       |                             |             
                                       |                             |             
                                  +-+--v----+                   +-+--v----+        
                                  | v1.2    |                   | v2.0    |        
                                  | apidocs |                   | apidocs |        
                                  +---------+                   +---------+        

```

ammmze:
Sorry...been busy with other stuff. We are now looking to upgrade to springfox, so I should be able to do this soon.

Issue #526, opened 2014-12-10
Support java Collection type


apkobozev:
Why swagger.models.Collections doesn't support java.util.Collection interface? 
Could it be implemented? 
At this time, if I have Collection<E> type in ResponseBody, it parsed poorly and has no info about collection element type. 

Thank you.

rlogiacco:
Please consider a `Collection` cannot be a `Map` as the latter doesn't inherit from the former.
This means a `Collection` can be safely mapped to a JSON `array` as much as `List` and `Set` do: any possible representation of `Collection` provides a `toArray` method which returns its array representation.

I belive this should be the default implementation and configuration should be used to override it, rather than the other way around as suggested by previous answer.

hestad:
Added the workaround without effect, so I had to change my implementation to using Set/List instead of Collection. Waiting for a patch.

rlogiacco:
@hestad if you think it might be worth it, you can try my fork

dilipkrish:
@hestad Are you using Collections as a generic or are you using it as a vanilla type. 

in the case that you're using it as a vanilla type, you'd need to use the ```directModelSubstitute``` option.

```java 
// substitutes Collection with List for all instances of collection 
plugin.directModelSubstitute(Collection.class, List.class);
```

In case you're using the generics version you need to create a new rule for that. Some examples [here](https://github.com/springdox/springdox/blob/master/swagger-models/src/main/java/com/mangofactory/swagger/models/configuration/SwaggerModelsConfiguration.java#L26-L34).

Explicitly this is the code you'd need to use
```java
...
AlternateTypeRule collectionRule 
       = newRule(
                    //replace Collection<T>
                    typeResolver.resolve(Collection.class, WildcardType.class), 
                    //with List<T>
                    typeResolver.resolve(List.class, WildcardType.class));
plugin.alternateTypeRules(collectionRule);
....
```

__NOTE:__ you need to use ```WildcardType``` from the package```com.mangofactory.swagger.models.alternates```

Let me know if you have any issues with getting that to work.

dilipkrish:
@rlogiacco let me know if that resolves your issue with collections as well. Swagger the specification, supports sets and lists distinctly, with a ```uniqueItems``` attribute that is set to true or false based on type. And having this library arbitrarily pick one based on your PR is not the right solution.

hestad:
[...]

PR #615, opened 2015-03-14
added support for java.util.Collection and recursive alternate type substitution

rlogiacco:
These changes fix issue #526 providing a representation for `java.util.Collection`, represented as lists and arrays.

At the same time I added recursive resolution of alternate types, so that if we have a return value of type `ResponseEntity<Resource<MyClass>>` and both `ResponseEntity` and `Resource` have been configured as `genericModelSubstitutes` then the documentation will report a response of type `MyClass`.

Without the recursive search, only the outer class would be replaced, resulting in a return value of type `Resource<MyClass>`.

### DIFF
--- a/swagger-models/src/main/java/com/mangofactory/swagger/models/Collections.java
+++ b/swagger-models/src/main/java/com/mangofactory/swagger/models/Collections.java
@@ -17,10 +17,12 @@ public class Collections {
   }
 
   public static ResolvedType collectionElementType(ResolvedType type) {
-    if (List.class.isAssignableFrom(type.getErasedType()) || Collection.class.isAssignableFrom(type.getErasedType())) {
-      return Collections.elementType(type, List.class);
-    } else if (Set.class.isAssignableFrom(type.getErasedType())) {
+    if (Set.class.isAssignableFrom(type.getErasedType())) {
       return Collections.elementType(type, Set.class);
+    } else if (List.class.isAssignableFrom(type.getErasedType())) {
+      return Collections.elementType(type, List.class);
+    } else if (Collection.class.isAssignableFrom(type.getErasedType())) {
+      return Collections.elementType(type, Collection.class);
     } else if (type.isArray()) {
       return type.getArrayElementType();
     } else {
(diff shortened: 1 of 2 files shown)


dilipkrish:
@rlogiacco Thanks for the PR! Since we're getting ready to release a 2.0-SNAPSHOT pretty soon, could you apply your fix to [that branch](https://github.com/springdox/springdox/tree/refactor/538/spike-to-seperate-models-and-dtos) instead. Keep in mind a LOT has changed.

Also preferably, with a few unit tests that validate your change. See guidelines [here](https://github.com/springdox/springdox/wiki/Contribution-guidelines). Especially because the change reg. recursive types, tho' tiny, seems risky to me; :fearful:  without having the safety net of a test.

Also, I'm :neutral_face: about the Collection being treated as an array. I'd rather have it as an extensibility via the substitution mechanisms already available.

rlogiacco:
@dilipkrish I believe I've already stated my point of view on the java.util.Collection thing, it's your choice to have it or not, but I don't see a reason not to have it. I tried to use the substitution mechanism as suggested in the mentioned issue, but it was not working in 1.0. I believe in convention over configuration, but it's up to the project leaders to decide which approach fits best for the project.
About the tests, it can be easily done, but I'm not going to implement that change on another code base if it's going to be rejected: please let me know if you want to merge that in or not.

dilipkrish:
@rlogiacco I'm still interested in seeing why ```Resource``` isn't substituted. Do you have a test case I could use?

rlogiacco:
```
@RequestMapping("/any/url")
public ResponseEntity<Resource<MyModel>> getMyModel() {
  return new ResponseEntity<Resource<MyModel>>(new Resource<MyModel>(new MyModel()), HttpStatus.OK);
}

@RequestMapping("/another/url")
public ResponseEntity<List<Resource<MyModel>>> searchMyModel() {
  List<Resource<MyModel>> result = new ArrayList<Resource<MyModel>>();
  result.add(new Resource<MyModel>(new MyModel()));
  return new ResponseEntity<List<Resource<MyModel>>>(result, HttpStatus.OK);
}

@RequestMapping("/any/other/url")
public @ResponseBody List<Resource<MyModel>> listMyModel() {
  List<Resource<MyModel>> result = new ArrayList<Resource<MyModel>>();
  result.add(new Resource<MyModel>(new MyModel()));
  return result;
}
```

I would expect the following response types:
* '/any/url' -> MyModel with model schema for MyModel
* '/another/url' -> List of MyModel with model schema of an array for MyModel
* '/any/other/url' -> List of MyModel with model schema of an array for MyModel

Instead I get:
* '/any/url' -> Resource<MyModel> with an empty model schema
* '/another/url' -> List of Resource<MyModel> with an empty model schema
* '/any/other/url' -> List of Resource<MyModel> with an empty model schema

dilipkrish:
@rlogiacco Closing this pr as the work around/ best practice is to provide an explicit ```AlternateTypeRule``` is to provide an explicit. Explanation of work around in #526.

PR #639, opened 2015-04-01
Update README.md

jordanjennings:
Remove commas in gradle dependencies, there should not be commas inside dependencies block


dilipkrish:
@jordanjennings Thanks!

Issue #659, opened 2015-04-09
Default Error Controller ("Basic Error Controller") - TRACE verb causes NPE in Swagger Codegen-2.1.2-M1 + duplicate nickname-problem


jfiala:
If you launch Swagger codegen (version com.wordnik:swagger-codegen-project:2.1.2M1) (https://github.com/swagger-api/swagger-codegen) on any API provided by Spring Swagger-MVC it will cause a NPE (regardless of language being used):

The error seems to be located in the Swagger Codegen project, not in Spring-Swagger-MVC

com.wordnik.swagger.codegen.Codegen -i http://localhost:8080/api-docs -l java

causes:
```java
java.lang.NullPointerException
	at io.swagger.parser.SwaggerCompatConverter.convert(SwaggerCompatConverter.java:455)
	at io.swagger.parser.SwaggerCompatConverter.read(SwaggerCompatConverter.java:111)
	at io.swagger.parser.SwaggerParser.read(SwaggerParser.java:34)
	at com.wordnik.swagger.codegen.Codegen.main(Codegen.java:88)
```

The SwaggerCompatConverter (located in swagger-compat-spec-parser-1.0.1.jar)  is referenced by SwaggerParser (located in swagger-parser-1.0.1.jar).

Quick fix:
Provide your own customer error converter:

``` Java
@Controller
@ApiIgnore
public class CustomErrorController implements ErrorController {

	private static final String PATH = "/error";

	@Override
	public String getErrorPath() {
		return PATH;
	}
	
	@RequestMapping(method=RequestMethod.GET, value=PATH, produces="application/json")
    @ResponseBody
    public Map<String, Object> handleGet(HttpServletRequest request) {

        Map<String, Object> map = new HashMap<String, Object>();
        map.put("status", request.getAttribute("javax.servlet.error.status_code"));
        map.put("reason", request.getAttribute("javax.servlet.error.message"));

        return map;
    }
	
	@RequestMapping(method=RequestMethod.POST, value=PATH, produces="application/json")
    @ResponseBody
    public Map<String, Object> handlePost(HttpServletRequest request) {

        Map<String, Object> map = new HashMap<String, Object>();
        map.put("status", request.getAttribute("javax.servlet.error.status_code"));
        map.put("reason", request.getAttribute("javax.servlet.error.message"));

        return map;
    }
	
	
}
```
Using this, you can generate the client sources using Swagger Codegen without errors.

[...]

PR #681, opened 2015-04-16
Adding ability to optionally populate operationId 

who:
I have added the ability to populate `operationId` in the rendered Swagger 2.0 output by reading the `nickname` parameter of `@ApiOperation`.  If no `nickname` is provided, no `operationId` will render.  Due to the issues discussed in https://github.com/springfox/springfox/issues/659, there is no support in this commit for automatically populating or generating `operationId`; it's completely up to the `@ApiOperation` annotation.

### DIFF
--- a/springfox-core/src/main/java/springfox/documentation/builders/OperationBuilder.java
+++ b/springfox-core/src/main/java/springfox/documentation/builders/OperationBuilder.java
@@ -54,7 +54,6 @@ public class OperationBuilder {
   private String deprecated;
   private boolean isHidden;
   private ModelRef responseModel;
-  private String nickname;
 
   /**
    * Updates the http method
@@ -233,20 +232,9 @@ public class OperationBuilder {
     return this;
   }
 
-  /**
-   * Sets the nickname for the operation
-   *
-   * @param nickname - a string for the nickname
-   * @return this
-   */
-  public OperationBuilder nickname(String nickname) {
-      this.nickname = defaultIfAbsent(nickname, this.nickname);
-      return this;
-  }
-
   public Operation build() {
     return new Operation(method, summary, notes, responseModel, uniqueId, position, tags, produces,
-        consumes, protocol, securityReferences, parameters, responseMessages, deprecated, isHidden, nickname);
+        consumes, protocol, securityReferences, parameters, responseMessages, deprecated, isHidden);
   }
 
   private Set<ResponseMessage> mergeResponseMessages(Set<ResponseMessage> responseMessages) {
--- a/springfox-core/src/main/java/springfox/documentation/builders/OperationBuilder.java
+++ b/springfox-core/src/main/java/springfox/documentation/builders/OperationBuilder.java
@@ -54,6 +54,7 @@ public class OperationBuilder {
   private String deprecated;
   private boolean isHidden;
   private ModelRef responseModel;
+  private String nickname;
 
   /**
    * Updates the http method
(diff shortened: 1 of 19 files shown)


springfox-core/src/main/java/springfox/documentation/builders/OperationBuilder.java
```diff
@@ -232,9 +233,20 @@ public OperationBuilder tags(Set<String> tags) {
...
+   */
+  public OperationBuilder nickname(String nickname) {
+      this.nickname = defaultIfAbsent(nickname, this.nickname);
+      return this;
+  }
+
   public Operation build() {
     return new Operation(method, summary, notes, responseModel, uniqueId, position, tags, produces,
-        consumes, protocol, securityReferences, parameters, responseMessages, deprecated, isHidden);
+        consumes, protocol, securityReferences, parameters, responseMessages, deprecated, isHidden, nickname);
```
dilipkrish: There is already a uniqueId in the operation model. I dont think we want to introduce swagger specific vocabulary into our internal models! If we change this I think this pr can be :merged: :+1:

dilipkrish:
@who also the coverage seems be down. I suspect the coverage will return back to normal if we remove the newly minted nickname attribute. Thanks for contributing back! :metal:

springfox-core/src/main/java/springfox/documentation/builders/OperationBuilder.java
who: @dilipkrish 

Currently, `uniqueId` is automatically generated based on HTTP verb and function name.

If I take the new `nickname` field out of the `OperationBuilder`, and instead use `uniqueId` as the `operationId` source in the `ServiceModelToSwagger2Mapper` class, which of the following should I do?
1. Do not generate a `uniqueId`, and populate it with `nickname` from `@ApiOperation`
2. Continue to generate a `uniqueId` as is but override it when a `nickname` is supplied in `@ApiOperation`
3. Other ?

dilipkrish: I prefer option2

adrianbk: @who thanks for contributing this!!

Referencing [the related issue](https://github.com/springfox/springfox/issues/659)

I would think that ideally we would have the following:
- Currently we do not include the operationId by default - this should remain.
- Provide a way for the user to configure and enable operationId generation via [Docket](https://github.com/springfox/springfox/blob/master/springfox-spring-web/src/main/java/springfox/documentation/spring/web/plugins/Docket.java#L63-63)
```java
. operationIdGenerator(someBeanOrFunctionThatPopulatesUniqueId())
```
- We should include a default `OperationIdGenerator` which generates a sensible id by default and would have the behaviour described in no.2 above (override it when a nickname is supplied in @ApiOperation). Users would enable it as follows:
```java
. operationIdGenerator(new OperationIdGenerator())
```

dilipkrish: @adrianbk I dont think we need to introduce an operation id generator concept. Thats what the plugin model is for, there is an operation builder plugin to enrich the operation model already. So l think the direction @who has taken would be the best approach. 

Just so we dont over engineer it, its probably fine to generate a operation id regardless. The extensibility mechanism allows folks to override the name if they wish.

who:
I have applied the first round of PR comments from @dilipkrish, removing nickname and instead using the existing uniqueId property to hold operationId. 

Also, code coverage should be passing now.

I will now address @adrianbk's feedback.

who:
Yikes - the last commit screwed up the whitespace for the test JSON files.  Fixing now.

springfox-core/src/main/java/springfox/documentation/builders/OperationBuilder.java
dilipkrish: What I mean by that is that its fine for us to generate an operationId in swagger 2.0 regardless. It is optional after-all and library users can change the name if they wish, either via @who's solution or by registering in their own ```OperationIdGenerator``` which is an ```OperationBuilderPlugin```

Also, I'd like to be a little more deliberate about introducing configuration options in Docket. For all intents and purposes, if we're doing it right, its going to be the source for all breaking changes.

I'd like to keep the extensibility hooks very narrow and as few as possible.

who:
Ok @dilipkrish, should be good to go.

dilipkrish:
:+1: @adrianbk you cool with this?

adrianbk:
@who would you mind adding some documentation to the readme? We don't want useful features like this getting lost of forgotten about :)

who:
@adrianbk Sure, I can do that.

who:
@adrianbk @dilipkrish 

Committed the readme changes here: https://github.com/who/springfox/commit/9e61e6b08ebfc4f7af632f57ef37e452ec6ebe96. Let me know if these are too verbose/wordy, and I can trim them down a bit.

PR: https://github.com/springfox/springfox/pull/684

PR #691, opened 2015-04-20
Update Swagger2Controller.java

sashevsky:
uri.getPort() returns -1 if port is not specified.
That's create an error in json spec:
{
    "swagger": "2.0",
    "info": {
        "version": "0.1",
        "title": "",
        "contact": {},
        "license": {}
    },
    "host": "localhost:-1",
    "basePath": "/rest",
......

Which swaggerui doesn't repair and it doesn't work properly.


dilipkrish:
Thanks for the PR! Could you squash your commits please? Also it would be great to have a test for this change.

sashevsky:
Hm, it my first commit to github :)
I'm trying to find "splashing" now.
I don't know groovy, that's why I can't create tests fast. Maybe I'll try.

dilipkrish:
@sashevsky congratulations on your first PR! Squashing is just a way to make the three commits you have above into one. It's just a way to reduce noise in your commit.

PR #724, opened 2015-05-05
Refactored so that static docs generation suppots multiple formats

dilipkrish:


dilipkrish:
@RobWin you want to review these changes?

RobWin:
Hi @dilipkrish 
The dispatcher looks good, but should you remove HTML and PDF from the list. The conversion should be done outside of the code via a Gradle or Maven Plugin.
I even prefer the solution to just create the Swagger JSON and even push the AsciiDoc conversion into a Gradle Plugin. This simplifies the incorporation  of examples and and hand-written descriptions into the documentation. See new V0.5.0 feature of swagger2markup to reduce the pollution of code with Swagger annotations.

dilipkrish:
By combining the tests you should be able to eliminate the ordering issue.

I added HTML and PDF because I thought thats the next progression within the scope of this module. Seems like we should be able to run the tests and generate either format no? 

@RobWin you're the expert here, so I'll let you weigh in on how you want to take it forward, but I'll state my understanding just to be clear. There are 2 approaches to solving the problem @ build time. 
1. One is to hook into the mock mvc framework and wire in our static docs handler to generate docs. That is what the staticdocs module is, right?
2. The second is to provide a maven/gradle plugin that will generate docs, similar to [swagger-maven-plugin](https://github.com/kongchen/swagger-maven-plugin). 

So are you suggesting we move away from 1 to 2 or are you thinking of creating a maven/gradle plugin to augment 1?

I'll leave this PR open, no hurry to get it in. This is just to modularize the code to eliminate duplication. Also gives a consistent entry point into the generator, rather than having to remember 2-3 implementations based on what you need to generate. This is not to add new features etc.

@adrianbk any inputs?

RobWin:
@dilipkrish @adrianbk 
I can't combine the tests, because I could have hundreds of unit tests which also create spring-restdocs HTTP request and response snippets.
Therefore the `convertSwaggerToMarkup` test must always run after or at the end of the unit test phase. 
This brought me to apprach 2.)

1. I created the PR so that springfox-staticdocs has a new ResultHandler which just creates the `swagger.json` file
2. I created a [Swagger2Markup Gradle Plugin](https://github.com/RobWin/swagger2markup-gradle-plugin) which converts the swagger.json into AsciiDoc or Markdown. It allows to chain the gradle plugins: test -> swagger2markup -> asciidoctor -> jar
The HTML or PDF documentation is then included into the jar file and can be served as static content.

dilipkrish:
That clears some confusion, what about maven users?

RobWin:
Maven users can still use approach 1.)

dilipkrish:
@RobWin merge or close that is the question? :8ball: Since much of the functionality is in your library, I'm inclined to even yank this module all together as it isn't providing any additional value over your awesome library. thoughts? 

/c @adrianbk

RobWin:
OK lets close it.

dilipkrish:
@robwin @adrianbk any ideas on how to take this forward?

dilipkrish:
closing this pr for now

PR #730, opened 2015-05-06
Bug when controller instance is proxied

mtalbot:
When adding hystrix commands are aop aspects around the services the bean type is a java proxy and so you loose the detailed response type. This causes that to fall back to using the methods return type directly.


dilipkrish:
Looks like a simple enough change. So does this mean all proxied objects will not have any models? Also PRs accompanied by tests that validate the change are much appreciated.

dilipkrish:
@mtalbot do you have the test, so that I can merge the PR?

dilipkrish:
:metal:

PR #760, opened 2015-05-28
Maintenance/759/build to not modify source controlled items

### LINKED ISSUE #759: Improve the build workflow
Currently there are a few manual steps that involve changing altering the code to switch between release mode and snapshot mode.

dilipkrish:
Resolves #759 

# OPEN FOR FEEDBACK 
### Merge after squashing

Introduced two concepts to be part of the CD build workflow
- First idea to use the git annotated tags as a way to generate version numbers, rather than a version.properties file. `git describe` and `git describe --exact-match` to the rescue. 

    Git already has an idea of generating serial predictable commit descriptions (build numbers) given an annotated tag for e.g. the currently the version according to git in this branch is `2.0.0-62-g8433c2b` it has the format `2.0.0-<count>-g<sha>`  where:
    - 2.0.0 is the annotated tag
    - count is the number of commits after the annotated tag (62 commits since 2.0.0!!)
    - sha is the commit sha. (8433c2b is the current sha)

- Second idea I introduced is a way to infer the snapshot build/ release build using the concept of a build number suffix. Rather than change the version decorator used.


adrianbk:
@dilipkrish looks good to me. No better way to test this than to merge it in and release!

Issue #675, opened 2015-04-14
Tags in @ApiOperation are not rendered


who:
Currently, as of Springfox master Swagger 2.0 snapshot, if I use tags inside an `@ApiOperation` block, they are not rendered in the api-doc output. Instead, the `value` attribute of the class-level `@Api` annotation is rendered. 

Given this annotation:

```
@RestController
@RequestMapping("/bookstore")
@Api(
    value = "this is the api value string in class level annotation",
    description = "API for book store")
public class BookStoreController {
...
    @RequestMapping(
        value = "/books", 
        method = RequestMethod.POST, 
        consumes = MediaType.APPLICATION_JSON_VALUE,       
        produces = MediaType.APPLICATION_JSON_VALUE)
    @ApiOperation(
        value = "Create a book",
        notes = "Creates a book so that it can be managed by this api.",
        response = Book.class,
        tags = {"foo", "bar"}
    )
    public Book createBook(
        @ApiParam(value = "Book object", required = true) @RequestBody @Valid Book book) {
        return bookService.createBook(book);
    }
```

This will be the output:

```
{
  ...
  "paths": {
    "/books": {
      "post": {
        "tags": [
          "this-is-the-api-value-string-in-class-level-annotation"
        ],
        "summary": "Creates a book",
        "description": "Creates a book so that it can be managed by this api.",
        "schemes": [],
        "consumes": [
          "application/json"
        ],
        "produces": [
          "application/json"
        ],
        "parameters": [
          {
            "in": "body",
            "name": "body",
            "description": "Book object",
            "required": true,
            "schema": {
              "$ref": "#/definitions/Book"
            }
          }
        ],
        "responses": {
          ...
        }
      }
    }
  }
  ...
}
```

Why are the "foo" and "bar" tags not included in the rendered output?

Also relevant, from https://github.com/swagger-api/swagger-spec/blob/master/versions/2.0.md#fixed-fields
> Not all tags that are used by the Operation Object must be declared.

dilipkrish:
@who because we haven't gotten to implementing the swagger 2.0 ONLY features :) We could sure use some help.

z0mb1ek:
fix this please =)

PR #772, opened 2015-06-06
#675 Initial approach to adding @Api(description) to the tags output

kevinconaway:
@dilipkrish I took a stab at fixing #675 

I refactored `ResourceGroupingStrategy` to use the `ResourceGroup` to get the resource description.  From my reading of the code base, this didn't need to be at the level of an individual handler method.

It looks like you could also apply the same logic to `getResourcePosition` or simply remove it all together because it didn't really look used in the production code outside of `ClassOrApiAnnotationResourceGrouping`


dilipkrish:
@kevinconaway thanks for the PR! A quick glance through it it seems like a good fix, however, I'll wait to merge it in for a couple of reasons:
- getResourceDescription doesn't belong in ResourceGroupingStrategy, So I was going to refactor that some so that the responsibility is separated.
- Secondly, a lot of this cruft is there support swagger 1.2 side by side and we're slowly reworking the internals of the swagger 1.2 codebase to be more general purpose; for e.g. swagger 2.0 doesn't have a concept of Listing references. 
- Lastly I'm :watch:  for swagger-core to change its package name to release 2.0.2 patch with the new swagger-core 1.5 release, so I don't want to pull in new changes before that happens.

kevinconaway:
Thanks for the quick response.  Any idea on a timeline for the above?

The top level API descriptions are the first thing end user sees when they view the Swagger docs so I'd like to be able to make a good impression when we release our API documentation.

dilipkrish:
@kevinconaway if you modify the PR such that it preserves backwards compatibility, i'll take merge the pr. ResourceGroupingStrategy is a public api, so I suspect there will be atleast a few custom implementations out in the wild. This change should not break those library consumers. Ideally we want to give atleast a (minor) version for the api consumers to catchup with deprecated apis.

I'd say re-instate the the deleted api and mark as deprecated.

dilipkrish:
@kevinconaway oops! we might have stepped on each others feet here... I just pushed an updated that touched a lot of files

kevinconaway:
I reverted the change to `ResourceGroupingStrategy` per your request, I understand and appreciate your concern for keeping the spi api stable.

kevinconaway:
I pulled in your changes locally and the project built fine without any conflicts. I pushed the latest to my branch so we should be OK there.

dilipkrish:
I squashed the merge commit, and I've pulled it in to master. Thanks for the pr :metal:

Issue #773, opened 2015-06-08
Swagger 2.0 - Cannot Have Blank Notes / Implementation Details


kevinconaway:
The default value for @ApiOperation(notes) is "" but Springfox substitutes the _operationName_ (which appears to be the method name) instead.

With this logic, it doesn't appear that there is a way to have the Implementation Details section blank like the default Petstore application does: http://petstore.swagger.io/

dilipkrish:
The idea was to surface that in the ui so library consumers will see that it needs to be overridden. Since its a plugin based system, you could override it back to blank.

kevinconaway:
IMO it shouldn't need to be overridden.  Take the petstore reference application for example, most of the API listings are plenty clear with just the HTTP verb, path and summary statement.

I'll test out the plugin approach.

dilipkrish:
I think just removing [this line](https://github.com/springfox/springfox/blob/master/springfox-spring-web/src/main/java/springfox/documentation/spring/web/readers/operation/DefaultOperationReader.java#L38) should do it :) and of course fixing the tests

kevinconaway:
Thanks, do you want me to submit a PR for that?  Or are you suggesting doing that via a custom plugin?

dilipkrish:
I think a plugin is probably overkill :) if you could do a pr that would be awesome

PR #777, opened 2015-06-08
#773 Don't provide a default value for operation notes/description.

kevinconaway:
@dilipkrish here is the PR for #773


dilipkrish:
@kevinconaway thanks for the pr! you're on a roll! :+1:

Issue #778, opened 2015-06-08
@RequestParam Field With Default Value Marked Required


kevinconaway:
If I have a `@RequestParam` field with a default value like:

`@RequestParam(value = "start", defaultValue = "0") int start`

Springfox will still mark this field as _required_ in the swagger.json.  Per the contract in [@RequestParam](http://docs.spring.io/spring/docs/current/javadoc-api/org/springframework/web/bind/annotation/RequestParam.html#defaultValue--):

> Supplying a default value implicitly sets required() to false.

It looks like the `ParameterRequiredReader` should also check if the `@RequestParam` has a _defaultValue_ on it and if so, not set the field to be required.

Happy to submit a PR for this if you wish.

dilipkrish:
sure, that would be awesome!

dilipkrish:
Fixed via #782

rawit:
I just ran in to this issue and were happy it seems to be fixed in 2.0.2. 

But I am using 2.6.0 so I am a little bit confused.

Mumi-aboaya:
If have the following example:
@RequestParam(value = "foo", defaultValue = "")
This will result in an required field even though it is not.
Just looked at the commit mentioned above.
There is a check for null AND not empty.
You should remove the "not empty" check as this is entirely up to the developer to decide.
For now, my fix is to use the following syntax:
@RequestParam(value = "foo", defaultValue = "", required = false)

dilipkrish:
@Mumi-aboaya By default RequestParam#required() is true. Just because it has a default value it doesnt make it not required.

Mumi-aboaya:
Please read the first post.
Also this is taken from the documentation:

> String org.springframework.web.bind.annotation.RequestParam.defaultValue() : 
> "\n\t\t\n\t\t\n\uE000\uE001\uE002\n\t\t\t\t\n"
> The default value to use as a fallback when the request parameter is not provided or has an empty value.  Supplying a default value **implicitly sets required to false**.

Basically the code should go from:
` boolean missingDefaultValue = ValueConstants.DEFAULT_NONE.equals(defaultValue) || (defaultValue == null || defaultValue.length() == 0);`

to:
` boolean missingDefaultValue = ValueConstants.DEFAULT_NONE.equals(defaultValue);`

Anything other than **ValueConstants.DEFAULT_NONE** should be considered set by code, hence required should be set to false.

dilipkrish:
[...]

PR #782, opened 2015-06-09
#778 Take @RequestParam(defaultValue) in to account when marking para…

kevinconaway:
…meter required.

@dilipkrish PR for #778


dilipkrish:
Thanks!! :metal:

PR #791, opened 2015-06-15
Updated documentation to show an example of springfox-staticdocs

RobWin:
Updated documentation to show an example of springfox-staticdocs.
Updated some broken URLs and code examples.


dilipkrish:
Thanks!

PR #831, opened 2015-07-03
Changes to allow for readOnly property in swagger 2

jfearon:
Changes added to process @ApiModelProperty(readonly=true).

Altered model tests, and propery spec test to ensure new flag was set correctly.

### DIFF
--- a/springfox-core/src/main/java/springfox/documentation/builders/ModelPropertyBuilder.java
+++ b/springfox-core/src/main/java/springfox/documentation/builders/ModelPropertyBuilder.java
@@ -32,6 +32,7 @@ public class ModelPropertyBuilder {
   private String qualifiedType;
   private int position;
   private Boolean required;
+  private Boolean readOnly;
   private String description;
   private AllowableValues allowableValues;
   private String name;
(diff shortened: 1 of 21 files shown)


dilipkrish:
Everything looks good! thanks for the PR! However, Im curious how this did not affect any of the contract tests :confused:

jfearon:
Do you have a preference for what model gets changed in the contract tests?  Or shall i just add a new variable to one of the classes?

dilipkrish:
Without changing any models, based on the pr I'd expect model properties to have readOnly properties to all be defaulted to false and consequently show up in the contract tests (for swagger2) as false. Wondering why that's not the case.

Regarding which model to add the behavior, any models in the feature demonstration service would be perfect.

jfearon:
Ah I see.  I assumed that this was why the required flag was Boolean not boolean and that some magic was happening, but I did not investigate further.  
I'll add a new instance variable to the Example class and change the tests.

dilipkrish:
I think the fact that you made it a `Boolean`  is perfect we don't want to change that. Probably use the `Optional.fromNullable().orNull()`  pattern instead of `Optional.fromNullable().or(false)`

dilipkrish:
Having said that, the logic for the required flag is probably not right too :) i can fix that

PR #847, opened 2015-07-15
Link to minified swagger-ui

ejain:
This reduces the amount of data that needs to be transferred by 2mb!


Lucas-C:
Directly related: https://github.com/api-platform/core/issues/3008

Issue #810, opened 2015-06-23
How to Correctly Model Array of Enums


kevinconaway:
How can I correctly model an array of enum values?

If I have the following:

```java
@ApiModel(value="A permission")
public enum Permission {
    @ApiModelProperty(value="Can read")
    read,
    @ApiModelProperty(value="Can write")
    write
}
```
and

```java
@ApiModel(value="Some resource"
public class MyResource {

    @ApiModelProperty(value="Permissions to apply")
     private Set<Permission> permissions;
}
```

I would expect that `MyResource` would correctly show the _permissions_ property to be an array with the _allowableValues_ set to the values of the enum.  However, the field is rendered as:

```javascript
permissions: {
    type: "array",
    description: "Permissions to apply",
    items: {
        type: "string"
    }
}
```

How should this be rendered?

dilipkrish:
Closing this as its a duplicate of #801. This is because response models do not support allowable values yet.

pbeaudouinexpedia:
My understanding of 06a33f0 is that allowable values are automatically documented based on the enum.

However, the example given by @kevinconaway still does nothing (no enum or enum field documentation).

Maven dependency tree:
> [INFO] +- io.springfox:springfox-swagger2:jar:2.9.0:compile
> [INFO] |  +- io.swagger:swagger-annotations:jar:1.5.18:compile
> [INFO] |  +- io.swagger:swagger-models:jar:1.5.18:compile
> [INFO] |  +- io.springfox:springfox-swagger-common:jar:2.9.0:compile
> [INFO] +- io.springfox:springfox-swagger-ui:jar:2.9.0:compile

Are we using the annotations incorrectly?

dilipkrish:
@pbeaudouinexpedia this is a very old issue. Could you perhaps give an example of what you're situation is? Your model and what you're seeing rendered in the spec?

pbeaudouinexpedia:
@dilipkrish I'm trying to document an enum with `@ApiModel` just like I am documenting a class. I also would like to document the enum values using `@ApiModelProperty` just like it works with class fields.

At the moment the workaround is to document the enum every time it is passed as a parameter, which isn't maintainable.

The example given in the first comment, https://github.com/springfox/springfox/issues/810#issue-90381700 is exactly what I am trying to do.

When I do this, the text entered in the `@ApiModel` and `@ApiModelProperty` annotations is not visible in Swagger UI. When I do the same on a class and class fields, the text is visible in Swagger UI.

[...]

Issue #801, opened 2015-06-17
Allowable Values for enumerations on response models do not render options


GitVhaos:
I`m not shure if this has to be adressed for swagger-spec or springfox-swagger:

When returning an Java-Enum it would be nice to see the "allowable Values" of this enum as possible results in the generated json-output. For using enums as Parameters this is already working, but not for using enums as response-object.

Where do I need to adress this feature?

<b>Example:</b>
```java
@RequestMapping(value = "/someResource", method = RequestMethod.GET)
  public ResponseEntity<SomeEnum> readResource()
```

dilipkrish:
This is probably already working. Just that the swagger-ui doesn't show it

GitVhaos:
Unfortunately not... In JSON-Sepc the enum is only shown as:

   "200":{
                  "description":"OK",
                  "schema":{
                     "type":"string"
                  }
               },

kevinconaway:
@dilipkrish WRT #810, I'm happy to work on this if you can provide direction on what the response should look like.  Is it not supported in the swagger 2.0 spec?  Or just not in springfox.

dilipkrish:
That would be awesome! For sure, swagger 1.2 doesn't support this notion. Swagger 2.0 does. 

Haven't really seen what the effort to support this is. If I remember right, it's a matter of having either the model ref expose allowable values or introduce this in the mapping layer for 2.0 such that it doesn't break the case where enums or list of enums are properties of input parameters. Will provide more details when I look at the code

kevinconaway:
@dilipkrish any idea on the above?  I'm still not sure what this should look like in the swagger json

dilipkrish:
In swagger 2: [this](https://github.com/springfox/springfox/blob/master/swagger-contract-tests/src/test/resources/contract/swagger2/declaration-business-service.json#L201-L234) would look like this

```json
        "/businesses/typeEcho": {
            "post": {
                "tags": [
                    "business-service"
                ],
                "summary": "businessTypeEcho",
                "operationId": "businessTypeEchoUsingPOST",
                "consumes": [
                    "application/json"
                ],
                "produces": [
                    "application/json"
                ],
                "parameters": [
                    {
                        "in": "body",
                        "name": "business",
                        "description": "business",
                        "required": true,
                        "schema": {
                            "type": "string"
                        }
                    }
                ],
                "responses": {
                    "200": {
                        "description": "OK",
                        "schema": {
                            "type": "string",
                            "enum": [ //<----
[...]

PR #849, opened 2015-07-15
#801 #810 Support for enums in response values and in schema objects

kevinconaway:
@dilipkrish Fixes for #801 and #810 based on our discussion in #801

### DIFF
--- a/springfox-schema/src/test/groovy/springfox/documentation/schema/property/field/FieldModelPropertyProviderSpec.groovy
+++ b/springfox-schema/src/test/groovy/springfox/documentation/schema/property/field/FieldModelPropertyProviderSpec.groovy
@@ -91,4 +91,30 @@ class FieldModelPropertyProviderSpec extends Specification {
       deSerializationPropNames == ['name']
 
   }
+
+  def "Handles enum collections" () {
+
+    given:
+    def genericNamingStrategy = new DefaultGenericTypeNamingStrategy()
+    Class typeToTest = collectionEnumType()
+    def typeResolver = new TypeResolver()
+    ResolvedType resolvedType = typeResolver.resolve(typeToTest)
+    ObjectMapper mapper = objectMapperThatUsesFields()
+    def namingStrategy = new ObjectMapperBeanPropertyNamingStrategy()
+    namingStrategy.onApplicationEvent(new ObjectMapperConfigured(this, mapper))
+    def propertyProvider = new FieldModelPropertyProvider(new FieldProvider(typeResolver),
+            namingStrategy, defaultSchemaPlugins(), Mock(TypeNameExtractor))
+    propertyProvider.objectMapper = mapper
+    def serializationPropNames = propertyProvider.propertiesFor(resolvedType, returnValue(resolvedType, DocumentationType.SWAGGER_12, alternateTypeProvider(), genericNamingStrategy)
+    )
+            .collect({it.name})
+    def deSerializationPropNames = propertyProvider.propertiesFor(resolvedType, inputParam(resolvedType, DocumentationType.SWAGGER_12, alternateTypeProvider(), genericNamingStrategy)
+    )
+            .collect({it.name})
+
+    expect:
+    serializationPropNames == ['exampleEnums']
+    deSerializationPropNames == ['exampleEnums']
+
+  }
 }
--- a/springfox-schema/src/test/groovy/springfox/documentation/schema/property/field/FieldModelPropertyProviderSpec.groovy
+++ b/springfox-schema/src/test/groovy/springfox/documentation/schema/property/field/FieldModelPropertyProviderSpec.groovy
@@ -24,6 +24,7 @@ import com.fasterxml.jackson.databind.ObjectMapper
 import spock.lang.Specification
 import springfox.documentation.schema.AlternateTypesSupport
 import springfox.documentation.schema.DefaultGenericTypeNamingStrategy
+import springfox.documentation.schema.ExampleEnum
 import springfox.documentation.schema.TypeNameExtractor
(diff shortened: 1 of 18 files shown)


springfox-schema/src/test/groovy/springfox/documentation/schema/property/field/FieldModelPropertyProviderSpec.groovy
```diff
@@ -91,4 +91,30 @@ class FieldModelPropertyProviderSpec extends Specification {
...
+    def serializationPropNames = propertyProvider.propertiesFor(resolvedType, returnValue(resolvedType, DocumentationType.SWAGGER_12, alternateTypeProvider(), genericNamingStrategy)
+    )
+            .collect({it.name})
+    def deSerializationPropNames = propertyProvider.propertiesFor(resolvedType, inputParam(resolvedType, DocumentationType.SWAGGER_12, alternateTypeProvider(), genericNamingStrategy)
+    )
+            .collect({it.name})
+
+    expect:
+    serializationPropNames == ['exampleEnums']
+    deSerializationPropNames == ['exampleEnums']
```
dilipkrish: We should probably assert that the properties are rendered correctly here. If its just the names we're checking we could parameterize the above test.

springfox-spring-web/src/main/java/springfox/documentation/spring/web/readers/operation/ModelRefs.java
```diff
@@ -58,7 +61,8 @@ public static ModelRef modelRef(ResolvedType resolved,
     if (Void.class.equals(resolved.getErasedType()) || Void.TYPE.equals(resolved.getErasedType())) {
       return new ModelRef("void");
     }
+    AllowableValues allowableValues = ResolvedTypes.allowableValues(resolved);
     String typeName = nameExtractor.typeName(fromParent(modelContext, resolved));
-    return new ModelRef(typeName);
+    return new ModelRef(typeName, null, allowableValues);
```
dilipkrish: May be provide a builder for ModelRef... :cry: for the null in the middle

dilipkrish:
:clap: minus a few nits. Thanks for the quick turn around!

dilipkrish:
might be able to squeeze this into 2.1.0 after all

kevinconaway:
Will update this tonight or tomorrow morningbased on your feedback

On Wednesday, July 15, 2015, Dilip Krishnan <notifications@github.com>
wrote:

> might be able to squeeze this into 2.1.0 after all
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/springfox/springfox/pull/849#issuecomment-121764635>.
>


-- 
Kevin Conaway
http://www.linkedin.com/pub/kevin-conaway/7/107/580/
https://github.com/kevinconaway

springfox-spring-web/src/main/java/springfox/documentation/spring/web/readers/operation/ModelRefs.java
kevinconaway: @dilipkrish I'm hesitant to add a builder at this point because then I'd feel compelled to change all references of `_new Model(...)_ to use the builder.  Perhaps this could be done as part of your planned refactoring?

I'll add a separate constructor to eliminate the null for now.  Will that work?

dilipkrish: 👍🏼

springfox-schema/src/test/groovy/springfox/documentation/schema/property/field/FieldModelPropertyProviderSpec.groovy
kevinconaway: Can you point me in the right direction about parameterizing the test?  I haven't used Spock before

dilipkrish: May be just verify that the properties itself rather than just the names, since this is new behavior

```groovy
def deSerializationProps = propertyProvider.propertiesFor(resolvedType, inputParam(resolvedType, DocumentationType.SWAGGER_12, alternateTypeProvider(), genericNamingStrategy))
expect:
     def prop = deSerializationProps.first()
     prop.name == "exampleEnums"
     .... //assert has allowable values etc.
    // same for serialization propererties
```

kevinconaway: Sure thing.  Is there a nice way with spock run the same _expect_ logic for both deserializationProps and serializationProps?

dilipkrish: Yeah sure, you could extract to a method and pass in the props.

springfox-schema/src/test/groovy/springfox/documentation/schema/property/field/FieldModelPropertyProviderSpec.groovy
```diff
@@ -91,4 +92,41 @@ class FieldModelPropertyProviderSpec extends Specification {
...
+      assert props.first().type.typeParameters.size() == 1
+      assert props.first().type.typeParameters.first().erasedType == ExampleEnum
+      assert props.first().allowableValues.values == ['ONE', 'TWO']
+      true
+    }
+
+    assertModelProperties(serializationProps)
+    assertModelProperties(deSerializationProps)
+  }
+
```
kevinconaway: @dilipkrish I updated the test case a bit, I'm not if there is a better way to accomplish this.

dilipkrish:
Thanks! @kevinconaway

PR #860, opened 2015-07-20
Updated swagger2markup version

RobWin:
Updated Swagger2Markup dependency from 0.7.0 to 0.7.1

Workaround: If the type of a BodyParameter is String and not a Model, the schema is null and lost. Therefore the fallback type of a BodyParameter is String now.


dilipkrish:
Thanks @RobWin!

Issue #845, opened 2015-07-14
[Question] Configuration for global Operation-Parameters


GitVhaos:
Im currently working on an internal API which uses two "parameters" for authentication. These two parameters have to be sent for each request to authenticate the user. The authentication mechanism is place in one central place, so no Spring Controller has these two parameters as its request-parameters. In Swagger (-UI) these parameters are now missing as Input Parameters.

In my opionion it`s no option to annotate every Controller with Swagger-Annotations declaring these parameters. Later developer will miss to configure it on new controllers. Also in a future api version I maybe want to change my authentication type without editing each Controller.

Is there an option (maybe via Docket) to configure parameters which will be added to each operation in swagger documentation?

I`m searching for something like this:

return new Docket(DocumentationType.SWAGGER_2)
    .select().build()
    .addGlobalRequestParameter(String.class, "user")
    .addGlobalRequestParameter(Integer.class, "department");

dilipkrish:
Thats a great idea. Would you like to take a shot at implementing it.

GitVhaos:
I will give it a try!

GitVhaos:
I`m currenty working on this feature. It seem`s that in "OperationBuilder.parameters()" the ParameterMergers constructor arguments are incorrect. "Destination" should be the existing parameters, while "source" should be the new parameters. Internal that`s not causing a bug but is kind of confusing. Can u please check this?

GitVhaos:
@dilipkrish I created a pull request for this feature.

GitVhaos:
This feature is already merged with  #861 and can be closed as duplicate.

dilipkrish:
This feature has been already implemented! via #861

jihlee:
This is great contribution :) I recently started using SpringFox and this feature was inevitable to use OAuth2 as authentication method for my system.

My question is, is it possible to exclude some specific methods for this configuration? All of my API methods should include 'Authorization' header except some authentication related APIs. 

At the moment, I could not find the way to achieve that so my authentication API also requires 'Authorization' header which is not nice.

GitVhaos:
Hi Jihun,
happy to hear that my contribution is also helpful for other people! 
Unfortunately I didn't had ur use case in our api. Cause we need authentication informations on every service.
You need to extend the current docket implementation to Support ur feature. I thought about a suggestion, but no "beautiful" way came to my mind...
[...]

Issue #755, opened 2015-05-26
o.s.data.domain.Pageable - automatically add @ApiImplicitParams?


steve-oakey:
Hi,

I have been looking for but haven't found a way to automatically the implicit parameters of the Pageable parameter. It would be great to have the page, size and sort parameters automatically added to the Swagger documentation. I have tried Docket#directModelSubstitute but this has not worked.

Any insight to this?

tonisostrat:
I had the same problem in a Spring project and from what I could find there is no "automagical" way to do this. You have two options:
* Add _@ApiIgnore_ to the *Pageable* and then specify each element as _@ApiImplicitParam_
* Change the method signature to accept each element of the *Pageable* as a separate argument

Support for this is debatable. *Pageable* is a Spring-specific construct and who's to say it will never change? Baking something like this into a general framework sounds counter-productive.

steve-oakey:
I agree, I am not suggesting that specific support for Pageable be written in. I was wondering if there was something in the Docket builder where I could configure my bean to say "if you see Pageable, use this class instead", where "this" class could be interpreted as request parameters (in the case of Pageable a class with page, size and sort fields).

dilipkrish:
This should be fixed via #845

steve-oakey:
Thanks!

zdila:
It seems #845 only adds global parameters to every operation. I've found no way to specify that the parameter should be used only when `Pageable` has been found. Am I missing something or is it really not implemented?

dilipkrish:
@zdila This was to only added implicit parameters to all operations. In your case it doesn't seem like the right solution. In your case, I would simply use an [alternate type rule](http://springfox.github.io/springfox/docs/current/#springfox-swagger2-with-spring-mvc-and-spring-boot) for `Pageable` (see **10**).

zdila:
OK, but description of this issue addresses exactly what I need but the provided solution can't be used for this.

I tried also the alternate type rule, but I can't use it to solve the problem.

What is needed is that for eg. `org.springframework.data.domain.Pageable` method parameter there should be created three `springfox.documentation.service.Parameter`s - for `page`, `offset` and `sort`, all with `paramType = "query"`.

dilipkrish:
Assuming that you're creating a concrete pageable object `MyRequest`, adding setters for your page/offset/sort etc. __OR__ *jackson*ifying it a little should do the trick

```java
class MyRequest extends AbstractPageRequest {
@JsonCreator
[...]

PR #861, opened 2015-07-20
Springfox #845: Configuration for global Operation-Parameters.

GitVhaos:
Added support for configuration of global operation parameters.

### DIFF
--- a/springfox-spi/src/main/java/springfox/documentation/spi/service/contexts/DocumentationContext.java
+++ b/springfox-spi/src/main/java/springfox/documentation/spi/service/contexts/DocumentationContext.java
@@ -31,6 +31,7 @@ import springfox.documentation.service.ApiDescription;
 import springfox.documentation.service.ApiInfo;
 import springfox.documentation.service.ApiListingReference;
 import springfox.documentation.service.Operation;
+import springfox.documentation.service.Parameter;
 import springfox.documentation.service.ResponseMessage;
 import springfox.documentation.service.SecurityScheme;
 import springfox.documentation.spi.DocumentationType;
@@ -51,6 +52,7 @@ public class DocumentationContext {
   private final AlternateTypeProvider alternateTypeProvider;
   private final Set<Class> ignorableParameterTypes;
   private final Map<RequestMethod, List<ResponseMessage>> globalResponseMessages;
+  private final List<Parameter> globalOperationParameters;
   private final ResourceGroupingStrategy resourceGroupingStrategy;
   private final PathProvider pathProvider;
   private final List<SecurityContext> securityContexts;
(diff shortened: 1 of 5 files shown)


dilipkrish:
@GitVhaos sorry I had forgotten about this! i'll take a look and merge it in if everything looks ok! Thanks very much for your pr! :+1:

dilipkrish:
@GitVhaos would be awesome if you could also update the docs :+1:

GitVhaos:
I`m new to gitHub... Where can I update the docs? ^^

dilipkrish:
Depends on what you're trying to document... but [here is the directory](https://github.com/springfox/springfox/tree/master/asciidoc) where the documentation lives.

thadc23:
Is there a way to hide/show these global parameters by `HttpMethod`? For example, I have a header param that I only want on every POST, PUT, and PATCH operation, but not on the others.  Before I give it a shot with a PR, I wanted to make sure it wasn't already possible and I was just missing it.

dilipkrish:
@thadc23 yes there is way, but at the moment you'd have to partition your apis into multiple dockets. So it will apply global parameters per docket. I believe there is a feature request to allow global parameters using a selector predicate. But it may be marked as a "Someday Maybe" feature. (I'll update this thread once I find it)

Let me know if you need any help working through it.

thadc23:
@dilipkrish Were you ever able to find the feature for global params using a selector predicate?  For a relatively small change and to implement filtering by method, I could add a List of HttpMethods to the Parameter and change the OperationParameterReader to filter the global list by the current operation's http method before adding the globals to the parameters for that operation.

To ensure backwards compatibility, I would add a second constructor to the Parameter that accepts a list of HttpMethods and have the current constructor call the new one with the List of all HttpMethods.

Thoughts?

dilipkrish:
I think there is a solution [described here](https://github.com/springfox/springfox/issues/755) that is preferable.

dilipkrish:
Basically creating a plugin that adds parameters conditionally.

thadc23:
Got it working using the plugin.  Thanks.

dilipkrish:
👍

avineshsachan:
Can we set GlobalOperationParameters in Docket Bean at run time as a authentication Token:

doc.select()                                  
          .apis(RequestHandlerSelectors.any())              
          .paths(PathSelectors.any())                        
          .build().globalOperationParameters(Lists.newArrayList(new ParameterBuilder()
          .name("sso-token")
          .description("Description of someGlobalParameter")
          .modelRef(new ModelRef("string"))
          .parameterType("header")
          .required(true).defaultValue("Run Tme Value get from SSO Service")
          .build()));

avineshsachan:
__

Can we set GlobalOperationParameters in Docket Bean at run time as a authentication Token:

docket.select()
.apis(RequestHandlerSelectors.any())
.paths(PathSelectors.any())
.build().globalOperationParameters(Lists.newArrayList(new ParameterBuilder()
.**name("sso-token")**
.description("Description of someGlobalParameter")
.modelRef(new ModelRef("string"))
.parameterType("header")
.required(true).defaultValue**_("Run Tme Value get from SSO Service")_**
.build()));

Issue #851, opened 2015-07-16
Sending API key to endpoints in request header


jgraniero:
Hi,

Some of the endpoints in my Spring Boot project are protected by a token.  The token must be present in the "Authorization" field in the header in order to access the endpoints

Is there any way to have swagger send  the API token with each request to the protected endpoints?  I tried using the example in the documentation that uses `ApiKey` as the security scheme.  This seems like exactly what I want, but I don't see the key getting sent in the header.

I'm happy to share some code if it helps, but I've tried setting `Docket.securitySchemes` and `Docket.securityContexts` with exactly the same values as the documentation and I don't see anything getting added to the request headers.

Is there something that needs to be configured in swagger-ui?  I'm pulling that in through `springfox-swagger-ui` and am using it as-is.

dilipkrish:
@jgraniero sharing code is :+1:

davidetrapani:
+1

I have a JWT protected API: the token is sent via a custom header (called X_AUTH_TOKEN).
It'd be great if the ApiKey specified in the security scheme was sent in the Request Header but unfortunately it doesn't seem so.

jgraniero:
@dilipkrish I will post some code by tomorrow, but suffice it to say that even the example from the documentation, which uses `ApiKey`, doesn't seem to be sending any key in the header.

simon-ras:
It seems that the api key does not get added due to a javascript error:

__Uncaught ReferenceError: log is not defined__

Looking at the swagger-ui repository, the log method is indeed defined: https://github.com/swagger-api/swagger-ui/blob/master/src/main/html/index.html#L78 but in SpringFox it is not: https://github.com/springfox/springfox/blob/master/springfox-swagger-ui/src/web/swagger-ui.html#L72

prabhat1790:
jgraniero : can you share your code after which you were able to get it working. Thanks

PR #862, opened 2015-07-20
#851 log method was not defined

simon-ras:


dilipkrish:
@simon-ras Thanks for identifying the issue and fixing it. Your change has been incorporated into 2.1.2!

Issue #870, opened 2015-07-22
"click to authenticate" button not clickable


Braamling:
I'm trying to configure OAuth with springfox (followup on #834). The button appears but is not clickable.

There is a small error icon next to the button: 
```html
<div class="auth">
      <span class="api-ic ic-error"></span>
         <div id="api_information_panel" style="top: 526px; left: 776px; display: none;">
         <div title="read and write">write</div>
     </div>
</div>
```

There is no error message, and the information panel appears somewhere else on the page. 

```java
public Docket swaggerSpringMvcPlugin()
{
     return new Docket( DocumentationType.SWAGGER_2 )
        .groupName( "test" )
        .select()
            .apis( not( withMethodAnnotation( PrivateAPI.class )))
            .paths( regex( ".*/rest/.*" ) )
            .build().apiInfo( new ApiInfoBuilder()
                .title( "test" )
                .description( "test" )
                .version( "0.1" )
                .build())
		.securitySchemes(securitySchemes())
		.securityContexts(securityContext()); 

	private class PasswordTokenRequestEndpoint extends TokenRequestEndpoint {

		private final String _username;
		private final String _password;

		public PasswordTokenRequestEndpoint(String url, String clientIdName,
				String clientSecretName, String username, String password) {

			super(url, clientIdName, clientSecretName);
			this._username = username;
			this._password = password;
		}

		@SuppressWarnings("unused")
		public String getUsername() {
			return _username;
		}

		@SuppressWarnings("unused")
		public String getPassword() {
			return _password;
		}
	}

	/**
	 * Class used for Oauth2 password grant type
	 */
	private class OAuth2PasswordCredentialsGrantType extends GrantType {

		@SuppressWarnings("unused")
		private final PasswordTokenRequestEndpoint _tokenRequestEndpoint;

		public OAuth2PasswordCredentialsGrantType(
				PasswordTokenRequestEndpoint tokenRequestEndpoint) {
			super("password");
			this._tokenRequestEndpoint = tokenRequestEndpoint;
		}
	}
	
	private List<SecurityContext> securityContext() {
		return Arrays.asList(SecurityContext.builder()
		.securityReferences(SecurityReferences())
		.build());
	}
	
	private List<SecurityScheme> securitySchemes() {
		ArrayList<SecurityScheme> authorizationTypes = new ArrayList<>();

		/* not sure why scopes are used for both authorization and authorization
		 * type
		 * API is not using the scopes for now
		 */
[...]

PR #903, opened 2015-08-10
fix for issue #870. "click to authenticate" button not clickable

sloppycoder:
fixed the issue of button not clickable and update o2c.html redirect path.


dilipkrish:
Thanks for the fix @sloppycoder!

PR #930, opened 2015-08-24
Updated swagger2markup version

RobWin:


dilipkrish:
Thanks @RobWin!! Any interesting features in this release?

RobWin:
A hook was added to preprocess the Swagger Model before it is converted into Markup. This allows to reorder model definitions or customize the Info section like basepath for different environments like prod or test.

PR #968, opened 2015-09-10
Fix generation of "schema" section for a "body" parameter to contain correct "type" & "format"

acourtneybrown:
I ran into a problem where the Swagger2 JSON document generated from a SpringBoot application with the following API method:

```java
  @RequestMapping(...)
  @ResponseStatus(...)
  public void setThreshold(..., @RequestBody int threshold) { ... }
```

would include the following for the `@RequestBody` parameter

```json
{
  "in": "body",
  "name": "threshold",
  "description": "threshold",
  "required": true,
  "schema": {
    "type": "int"
  }
}
```

This JSON has `"int"` for the `"type"` property, when it should be `"integer"`, with `"format": "int32"`.

This pull request updates `ParameterMapper` to ensure that it uses `io.swagger.models.properties.Property` for the `"schema"` `"type"` and `"format"` information, similar to what is done for collections and maps in that class.


dilipkrish:
Thank you for catching that! and for the pr! :beers:

PR #970, opened 2015-09-10
Proper handling of @ResponseStatus with default reason().

acourtneybrown:
- Javadoc for @ResponseStatus.reason() says that if no value is given, the
  default message from HttpStatus will be used.


dilipkrish:
Thanks!!!

PR #993, opened 2015-10-02
Fix handling of RequestMapping that include regex with quantifier(s)

### LINKED ISSUE #991: Extra closing curly brace on endpoints defined with a regex containing a quantifier
When using Springfox with endpoints containing URI path parameters with a defined regex containing a quantifier, an extra closing curly brace is inserted at the end of the URI parameter. The approach being used is described here: http://docs.spring.io/spring/docs/3.2.x/spring-framework-reference/html/mvc.html#mvc-ann-requestmapping-uri-templates-regex. 

For example, with the following annotation:

`@RequestMapping(value = "/scored/{path1}/{id:\\d{3}}/{code:\\d{5}}", produces = APPLICATION_JSON_VALUE, method = RequestMethod.GET)`

This is the generated path in the API spec:

`/scored/{path1}/{id}}/{code}}`

This also causes the Swagger UI to function improperly when testing out endpoints.

My guess is this is related to the curly braces on the quantifier that we're using in the regex.

If this is actually an issue with Swagger, please let me know and I can raise the defect with them.

erikthered:
Fixes #991

### DIFF
--- a/springfox-spring-web/src/main/java/springfox/documentation/spring/web/paths/Paths.java
+++ b/springfox-spring-web/src/main/java/springfox/documentation/spring/web/paths/Paths.java
@@ -88,7 +88,7 @@ public class Paths {
   public static String sanitizeRequestMappingPattern(String requestMappingPattern) {
     String result = requestMappingPattern;
     //remove regex portion '/{businessId:\\w+}'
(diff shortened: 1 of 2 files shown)


dilipkrish:
@erikthered thank you for being thorough!! :metal:

PR #1005, opened 2015-10-12
Springfox #845: Configuration for global Operation-Parameters. (Documentation)

GitVhaos:
Now (better late than never) the docu on issue #845


dilipkrish:
@GitVhaos thanks! :metal:

Issue #998, opened 2015-10-08
@ApiModelProperty example field is ignored


TheSneak:
@ApiModelProperty(value = "Date as string", example = "2015-10-07T17:00:52.950"
    private String test;

Annotating a property and setting the example field has no effect:

    "test": {
        "type": "string",
        "description": "Date as string"
    },

dilipkrish:
This hasn't been implemented. Would you like to try your hand at submitting a PR?

jrhowell:
I'd also find it very useful to have implemented. Is there any way that someone who wrote any of the other fields (like 'isRequired') could take a look at this or give hints on how to add this?
I've had a quick scan through the code (wondering whether I could fix it) and found where ApiModelProperty and ApiModelPropertyPropertyBuilder need to support 'example' but where does the actual production of swagger output happen?

dilipkrish:
You'e exactly on the right track ... the final piece is after adding example to the internal models and their builders is to `map` the internal model property using the `ModelMapper` [here](https://github.com/springfox/springfox/blob/master/springfox-swagger2/src/main/java/springfox/documentation/swagger2/mappers/ModelMapper.java#L103-L126). Would love to see a PR!

jrhowell:
I've got some code that I think solves the problem. But I don't have access to be able to push my code up to the springfox repo. Can you help with that? Thanks.

dilipkrish:
By access you mean access to submitting a PR? Here is a [simple guide](https://help.github.com/articles/using-pull-requests/)

jrhowell:
Thanks for the help Dilip. I've put in a pull request and it's now passing the test coverage of 96% too.

dilipkrish:
@jrhowell thanks jhowell... I'll merge it in manually, since it has a few changes to the imports that are superfluous (reordered). Apologies for not catching that earlier, but its a good opportunity to add to the coding styles document!

jrhowell:
Sorry about that. The code ended up going through an IDE that automatically re-ordered the imports. It also changed all sorts of other things, like indentation in the Copyright header to each file but I rolled those back. I didn't think to reorder the imports as well.

Out of curiosity I see that 2.3.0 is the next version, but there's no due date. Any rough idea of when that might be available? Thanks.

dilipkrish:
@jrhowell it should've been 2-3 weeks ago, but I have been a lil busy the last few weeks! :) hopefully will get it out end of this week

dilipkrish:
[...]

PR #1041, opened 2015-11-06
Added a fix for 'example' field in @ApiModelProperty annotation

jrhowell:
Things like @ApiModelProperty(value="value description", example="example value") didn't produce the example in the Swagger JSON output. Now it should.


jrhowell:
This is for issue: https://github.com/springfox/springfox/issues/998

dilipkrish:
@jrhowell sorry i'll get this merged soon, just back from a :bird: break!

PR #1042, opened 2015-11-06
Fixing logging error

AndreTurgeon:
log.info("{0}", "hi") -> "{0}"
log.info("{}", "hi" -> "hi"
I believe the second form is preferred in this case.


dilipkrish:
Hmm seems like that should be working now! thanks for the fix! :+1:

PR #1045, opened 2015-11-09
Fix issue #767, duplicate @JsonProperty

glarfs:
There is a duplicated @JsonProperty on the file, could you integrate it? Please

### DIFF
--- a/springfox-swagger-common/src/main/java/springfox/documentation/swagger/web/SecurityConfiguration.java
+++ b/springfox-swagger-common/src/main/java/springfox/documentation/swagger/web/SecurityConfiguration.java
@@ -71,7 +71,7 @@ public class SecurityConfiguration {
     return clientSecret;
   }
 
-  @JsonProperty("apiKey")
+  @JsonProperty("scopeSeparator")
   public String scopeSeparator() {
     return scopeSeparator;
   }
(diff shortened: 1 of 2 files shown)


springfox-swagger-common/src/main/java/springfox/documentation/swagger/web/SecurityConfiguration.java
```diff
@@ -71,7 +71,7 @@ public String getClientSecret() {
     return clientSecret;
   }
 
-  @JsonProperty("apiKey")
```
glarfs: apiKey duplicated

dilipkrish:
@glarfs gaah! thats some copy paste error right there! Thanks for catching that... looks like it has failing tests tho'

glarfs:
The test on my local branch(which is pointing to master) without this commit and it fails on the same point, it's strange. Seems that the code which is failing test is another commit.

n0mer:
@glarfs @dilipkrish  there are definitely 2 failing tests

image

glarfs:
There where an issue with tests, Now the tests are fixed

dilipkrish:
Thanks! :+1:

glarfs:
Hi, this fix could be applied also on 2.2.x branch, When are you releasing a new version with this fix?

PR #1078, opened 2015-12-02
fix async bug

rockytriton:
not sure if this is how you want it fixed... but this works when i use it...


dilipkrish:
:+1: thanks so much!. Also would you mind preserving the spaces. That way its clear whats changed. We generally use 2 spaces for indents.

rockytriton:
sure as soon as i figure out how to fix the pull requests, i already fixed the tabs->spaces

dilipkrish:
It appears its using the right indentation now :open_mouth:  Let me review and pull it in

Issue #1109, opened 2015-12-22
v2/api-docs does not produce application/json


vmarusic:
In cases where jackson-dataformat-xml is used in the Spring project (e.g. Spring Boot Actuator gets pulled which is common case) hitting the v2/api-docs endpoint with the browser will result in
response header:

**Content-type: application/xml;charset=UTF-8**

due to the fact that the browsers have  response header:

**Accept:text/html,application/xhtml+xml,application/xml;q=0.9,image/webp,*/*;q=0.8**

To fix this `getDocumentation` mapping of `Swagger2Controller.java` should be extended with the 
**produces={"application/json"}**

Thanks
V

dilipkrish:
Closing this issue as it was merged in and released

PR #1110, opened 2015-12-22
v2/api-docs does not produce application/json

vmarusic:
See #1109 for more details


dilipkrish:
@vmarusic thanks!! :+1:

Issue #1126, opened 2016-01-06
Race condition and crash on Swagger UI startup


chornyi:
When I load Swagger UI by going to `http://<host>/swagger-ui.html`, about 10% of the time I get an empty screen and this error in JS console:

	Uncaught TypeError: Cannot read property 'headerView' of undefined
		(anonymous function)  @ springfox.js:22
		p.event.dispatch @ jquery-1.8.0.min.js:2
		g.handle.h @ jquery-1.8.0.min.js:2
		p.event.trigger @ jquery-1.8.0.min.js:2
		(anonymous function) @ jquery-1.8.0.min.js:2
		p.extend.each @ jquery-1.8.0.min.js:2
		p.fn.p.each @ jquery-1.8.0.min.js:2
		p.fn.extend.trigger @ jquery-1.8.0.min.js:2
		p.fn.(anonymous function) @ jquery-1.8.0.min.js:2
		(anonymous function) @ springfox.js:50
		k @ jquery-1.8.0.min.js:2
		l.fireWith @ jquery-1.8.0.min.js:2
		y @ jquery-1.8.0.min.js:2d @ jquery-1.8.0.min.js:2

This happens both on Chrome and Firefox and I see it happen more often when I access the page over the network as opposed to localhost.

I interpret this as a race condition between the Swagger UI application initialization and the Springfox code to set the base URI when the `$('#select_baseUrl').change()` event is fired during the `$(document).ready()` callback. Sometimes these callbacks are fired before `window.swaggerUi` is set.

I'm talking about these two places in code:
https://github.com/springfox/springfox/blob/master/springfox-swagger-ui/src/web/swagger-ui.html#L30
https://github.com/springfox/springfox/blob/master/springfox-swagger-ui/src/web/js/springfox.js#L35

Is there a reason why this initialization happens in two functions that are executed in parallel?

Version `2.3.1`

dilipkrish:
The idea behind the two initializations was to preserve swagger-ui.html so that its easy to take in updates  from the mother ship. Open to ideas on how best to do that.

chornyi:
Let me see if I can figure out how to build this gradle thing and make the initialization sequential.

dilipkrish:
Let me know if you need help with that.. but it fairly straight forward

```bash
./gradlew :springfox-swagger-ui:build
```
will build your project. All it does is downloads the latest version of swagger-ui from the repo and repackages it as a webjar along with springfox customizations.

Thanks for looking into it! :+1:

[...]

PR #1127, opened 2016-01-06
Setup base URL after Swagger UI is initialized #1126

chornyi:


dilipkrish:
:metal: Thank you for doing this!... will test it out in a bit and merge it!

dilipkrish:
@chornyi thanks! :metal:

PR #1128, opened 2016-01-07
Fix Nullpointer exception when spring framework failed to get paramter name

MinosPong:
When using JAVA8 without enable -parameters in javac, spring will failed to obtain the parameter name . see [http://stackoverflow.com/questions/21455403/how-to-get-method-parameter-names-in-java-8-using-reflection]

 It causes springfox throw nullpointerexception.

This fix allow springfox using @ApiParam by default and prevent the nullpointerexception when Spring failed to obtain parameter name.


dilipkrish:
Thanks! :metal:

PR #1152, opened 2016-01-22
Add Polymorphism Support

Fyro-Ing:
I've added "Polymorphism support" (discriminator on parent, and allOf on child(s)) with springfox

```
{
  "definitions": {
    "Pet": {
      "type": "object",
      "discriminator": "petType",
      "properties": {
        "name": {
          "type": "string"
        },
        "petType": {
          "type": "string"
        }/
      },
      "required": [
        "name",
        "petType"
      ]
    },
    "Cat": {
      "description": "A representation of a cat",
      "allOf": [
        {
          "$ref": "#/definitions/Pet"
        },
        {
          "type": "object",
          "properties": {
            "huntingSkill": {
              "type": "string",
              "description": "The measured skill for hunting",
              "default": "lazy",
              "enum": [
                "clueless",
                "lazy",
                "adventurous",
                "aggressive"
              ]
            }
          },
          "required": [
            "huntingSkill"
          ]
        }
      ]
    },
    "Dog": {
      "description": "A representation of a dog",
      "allOf": [
        {
          "$ref": "#/definitions/Pet"
        },
        {
          "type": "object",
          "properties": {
            "packSize": {
              "type": "integer",
              "format": "int32",
              "description": "the size of the pack the dog is from",
              "default": 0,
              "minimum": 0
            }
          },
          "required": [
            "packSize"
          ]
        }
      ]
    }
  }
}
```

On java models : 
Parent -> @ApiModel(discriminator="id")
Child -> @ApiModel(parent = Parent.class)

see https://github.com/Fyro-Ing/springfox/commit/43923f583dc85e009f11f623655624a903042abb

### DIFF
--- a/springfox-core/src/main/java/springfox/documentation/schema/Model.java
+++ b/springfox-core/src/main/java/springfox/documentation/schema/Model.java
@@ -36,6 +36,7 @@ public class Model {
   private final String discriminator;
   private final List<String> subTypes;
   private final String example;
+  private final Model parent;
 
   public Model(
       String id,
@@ -47,7 +48,8 @@ public class Model {
       String baseModel,
       String discriminator,
       List<String> subTypes,
-      String example) {
+      String example,
+      Model parent) {
 
     this.id = id;
     this.name = name;
@@ -59,6 +61,7 @@ public class Model {
     this.discriminator = discriminator;
     this.subTypes = subTypes;
     this.example = example;
+    this.parent = parent;
   }
 
   public String getId() {
@@ -99,5 +102,9 @@ public class Model {
 
   public String getExample() {
     return example;
+  }
+
+  public Model getParent() {
+    return parent;
   }
 }
--- a/springfox-swagger-common/src/main/java/springfox/documentation/swagger/schema/ApiModelBuilder.java
+++ b/springfox-swagger-common/src/main/java/springfox/documentation/swagger/schema/ApiModelBuilder.java
@@ -20,11 +20,16 @@
 package springfox.documentation.swagger.schema;
 
 import com.fasterxml.classmate.TypeResolver;
+import com.google.common.base.Optional;
 import io.swagger.annotations.ApiModel;
 import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.beans.factory.annotation.Qualifier;
 import org.springframework.core.annotation.AnnotationUtils;
 import org.springframework.core.annotation.Order;
 import org.springframework.stereotype.Component;
+import org.springframework.util.StringUtils;
+import springfox.documentation.schema.Model;
+import springfox.documentation.schema.ModelProvider;
 import springfox.documentation.spi.DocumentationType;
 import springfox.documentation.spi.schema.ModelBuilderPlugin;
 import springfox.documentation.spi.schema.contexts.ModelContext;
@@ -36,10 +41,13 @@ import static springfox.documentation.swagger.common.SwaggerPluginSupport.*;
 @Order(SwaggerPluginSupport.SWAGGER_PLUGIN_ORDER)
 public class ApiModelBuilder implements ModelBuilderPlugin {
   private final TypeResolver typeResolver;
+  private final ModelProvider modelProvider;
 
   @Autowired
-  public ApiModelBuilder(TypeResolver typeResolver) {
+  public ApiModelBuilder(TypeResolver typeResolver,
+                         @Qualifier("cachedModels") ModelProvider modelProvider) {
     this.typeResolver = typeResolver;
+    this.modelProvider = modelProvider;
   }
 
 
@@ -48,6 +56,19 @@ public class ApiModelBuilder implements ModelBuilderPlugin {
     ApiModel annotation = AnnotationUtils.findAnnotation(forClass(context), ApiModel.class);
     if (annotation != null) {
       context.getBuilder().description(annotation.description());
+      if(StringUtils.hasText(annotation.discriminator())){
+        context.getBuilder().discriminator(annotation.discriminator());
+      }
+
+      // default parent is Void
+      if (!annotation.parent().isAssignableFrom(Void.class)) {
+        Optional<Model> parentModel = modelProvider.modelFor(ModelContext.fromParent(context, this.typeResolver.resolve(annotation.parent())));
+
+        // parent model exist because @ApiModel not required on model
+        // if(parentModel.isPresent()) {
+          context.getBuilder().parent(parentModel.get());
+        // }
+      }
     }
   }
 
--- a/springfox-core/src/main/java/springfox/documentation/builders/ModelBuilder.java
+++ b/springfox-core/src/main/java/springfox/documentation/builders/ModelBuilder.java
@@ -39,6 +39,7 @@ public class ModelBuilder {
   private String discriminator;
   private ResolvedType modelType;
   private String example;
+  private Model parent;
 
   private Map<String, ModelProperty> properties = newHashMap();
   private List<String> subTypes = newArrayList();
@@ -155,8 +156,19 @@ public class ModelBuilder {
     return this;
   }
 
+  /**
+   * Represents the parent type information with full fidelity of generics
+   *
+   * @param parent - resolved type that represents the parent model
+   * @return this
+   */
+  public ModelBuilder parent(Model parent) {
+    this.parent = defaultIfAbsent(parent, this.parent);
+    return this;
+  }
+
   public Model build() {
     return new Model(id, name, modelType, qualifiedType, properties, description, baseModel, discriminator, subTypes,
-        example);
+        example, parent);
   }
 }
--- a/springfox-core/src/test/groovy/springfox/documentation/builders/ModelBuilderSpec.groovy
+++ b/springfox-core/src/test/groovy/springfox/documentation/builders/ModelBuilderSpec.groovy
@@ -21,6 +21,7 @@ package springfox.documentation.builders
 
 import com.fasterxml.classmate.TypeResolver
 import spock.lang.Specification
+import springfox.documentation.schema.Model
 import springfox.documentation.schema.ModelProperty
 
 class ModelBuilderSpec extends Specification {
@@ -46,6 +47,7 @@ class ModelBuilderSpec extends Specification {
       'subTypes'      | ["String"]                            | 'subTypes'
       'properties'    | [p1: Mock(ModelProperty)]             | 'properties'
       'example'       | 'example1'                            | 'example'
+      'parent'        | Mock(Model)                           | 'parent'
   }
 
   def "Setting builder properties to null values preserves existing values"() {
@@ -71,5 +73,6 @@ class ModelBuilderSpec extends Specification {
       'subTypes'      | ["String"]                            | 'subTypes'
       'properties'    | [p1: Mock(ModelProperty)]             | 'properties'
       'example'       | 'example1'                            | 'example'
+      'parent'        | Mock(Model)                           | 'parent'
   }
 }
--- a/springfox-schema/src/test/groovy/springfox/documentation/schema/mixins/TypesForTestingSupport.groovy
+++ b/springfox-schema/src/test/groovy/springfox/documentation/schema/mixins/TypesForTestingSupport.groovy
@@ -200,4 +200,12 @@ class TypesForTestingSupport {
   def ResolvedType typeForTestingPropertyPositions() {
     resolver.resolve(TypeForTestingPropertyPositions)
   }
+
+  def ResolvedType typeForInheritedComplexType() {
+    resolver.resolve(InheritedComplexType)
+  }
+
+  def ResolvedType typeForComplexType() {
+    resolver.resolve(ComplexType)
+  }
 }
--- a/springfox-spring-web/src/main/java/springfox/documentation/spring/web/scanners/ApiModelReader.java
+++ b/springfox-spring-web/src/main/java/springfox/documentation/spring/web/scanners/ApiModelReader.java
@@ -113,6 +113,7 @@ public class ApiModelReader  {
                 .baseModel(targetModelValue.getBaseModel())
                 .discriminator(targetModelValue.getDiscriminator())
                 .subTypes(targetModelValue.getSubTypes())
+                .parent(targetModelValue.getParent())
                 .build();
 
         target.put(sourceModelKey, mergedModel);
--- a/springfox-swagger-common/src/test/groovy/springfox/documentation/swagger/mixins/SwaggerPluginsSupport.groovy
+++ b/springfox-swagger-common/src/test/groovy/springfox/documentation/swagger/mixins/SwaggerPluginsSupport.groovy
@@ -21,6 +21,8 @@ package springfox.documentation.swagger.mixins
 
 import com.fasterxml.classmate.TypeResolver
 import org.springframework.plugin.core.PluginRegistry
+import springfox.documentation.schema.mixins.ModelProviderSupport
+import springfox.documentation.schema.mixins.SchemaPluginsSupport
 import springfox.documentation.schema.plugins.SchemaPluginsManager
 import springfox.documentation.spi.DocumentationType
 import springfox.documentation.spi.schema.ModelBuilderPlugin
@@ -42,13 +44,14 @@ import static com.google.common.collect.Lists.*
 import static org.springframework.plugin.core.OrderAwarePluginRegistry.*
 
 @SuppressWarnings("GrMethodMayBeStatic")
+@Mixin([ModelProviderSupport, SchemaPluginsSupport])
 class SwaggerPluginsSupport {
   SchemaPluginsManager swaggerSchemaPlugins() {
     PluginRegistry<ModelPropertyBuilderPlugin, DocumentationType> propRegistry =
         create(newArrayList(new ApiModelPropertyPropertyBuilder()))
 
     PluginRegistry<ModelBuilderPlugin, DocumentationType> modelRegistry =
-        create(newArrayList(new ApiModelBuilder(new TypeResolver())))
+        create(newArrayList(new ApiModelBuilder(new TypeResolver(), defaultModelProvider())))
 
     new SchemaPluginsManager(propRegistry, modelRegistry)
   }
--- a/springfox-swagger-common/src/test/groovy/springfox/documentation/swagger/schema/ApiModelBuilderSpec.groovy
+++ b/springfox-swagger-common/src/test/groovy/springfox/documentation/swagger/schema/ApiModelBuilderSpec.groovy
@@ -3,19 +3,18 @@ package springfox.documentation.swagger.schema
 import com.fasterxml.classmate.TypeResolver
 import io.swagger.annotations.ApiModel
 import spock.lang.Shared
-import spock.lang.Specification
 import springfox.documentation.schema.DefaultGenericTypeNamingStrategy
-import springfox.documentation.spi.DocumentationType
+import springfox.documentation.schema.SchemaSpecification
 import springfox.documentation.spi.schema.AlternateTypeProvider
 import springfox.documentation.spi.schema.contexts.ModelContext
 
-class ApiModelBuilderSpec extends Specification {
+class ApiModelBuilderSpec extends SchemaSpecification {
   @Shared def resolver = new TypeResolver()
 
   def "Api model builder parses ApiModel annotation as expected" () {
     given:
-      ApiModelBuilder sut = new ApiModelBuilder(resolver)
-      ModelContext context = ModelContext.inputParam(type, DocumentationType.SWAGGER_12,
+      ApiModelBuilder sut = new ApiModelBuilder(resolver, modelProvider)
+      ModelContext context = ModelContext.inputParam(type, documentationType,
           new AlternateTypeProvider([]), new DefaultGenericTypeNamingStrategy())
     when:
       sut.apply(context)
@@ -31,5 +30,64 @@ class ApiModelBuilderSpec extends Specification {
   @ApiModel(description = "description")
   class AnnotatedTest {
 
+  }
+
+  def "Api model builder parses ApiModel annotation discriminator as expected" () {
+    given:
+      ApiModelBuilder sut = new ApiModelBuilder(resolver, modelProvider)
+      ModelContext context = ModelContext.inputParam(type, documentationType,
+          new AlternateTypeProvider([]), new DefaultGenericTypeNamingStrategy())
+    when:
+      sut.apply(context)
+    then:
+      context.builder.build().discriminator == expected
+    where:
+      type                | expected
+      String              | null
+      DiscriminatorTest   | "discriminator"
+
+  }
+
+  @ApiModel(discriminator = "discriminator")
+  class DiscriminatorTest {
+
+  }
+
+  def "Api model builder parses ApiModel annotation parent as expected" () {
+    given:
+      ApiModelBuilder sut = new ApiModelBuilder(resolver, modelProvider)
+      ModelContext context = ModelContext.inputParam(type, documentationType,
+          new AlternateTypeProvider([]), new DefaultGenericTypeNamingStrategy())
+    when:
+      sut.apply(context)
+    and:
+      def enriched = context.builder.build()
+    then:
+      enriched.parent.name == expected
+    where:
+      type                | expected
+      ParentTest          | "DiscriminatorTest"
+
+  }
+
+  @ApiModel(parent = DiscriminatorTest.class)
+  class ParentTest {
+
+  }
+
+  def "Api model builder parses ApiModel annotation without parent" () {
+    given:
+      ApiModelBuilder sut = new ApiModelBuilder(resolver, modelProvider)
+      ModelContext context = ModelContext.inputParam(type, documentationType,
+        new AlternateTypeProvider([]), new DefaultGenericTypeNamingStrategy())
+    when:
+      sut.apply(context)
+    and:
+      def enriched = context.builder.build()
+    then:
+      enriched.parent == expected
+    where:
+      type                | expected
+      AnnotatedTest       | null
   }
 }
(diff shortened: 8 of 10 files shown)


dilipkrish:
@Fyro-Ing This pr works but I'd like to re-work some aspects of it. Comments inlined

springfox-core/src/main/java/springfox/documentation/schema/Model.java
```diff
@@ -47,7 +48,8 @@ public Model(
       String baseModel,
       String discriminator,
       List<String> subTypes,
-      String example) {
+      String example,
+      Model parent) {
```
dilipkrish: I'd like tho use `ModelRef` here directly assuming some other plugin is responsible to dig out the `Model` from the `@ApiModel#parent()`. Typically by modifying the `ModelDependencyProvider` i.e. Parent => is a dependency of => Subclass

Secondly, we're anyways turning around and resolving it to a RefModel in the mapper.

springfox-swagger-common/src/main/java/springfox/documentation/swagger/schema/ApiModelBuilder.java
```diff
@@ -36,10 +41,13 @@
 @Order(SwaggerPluginSupport.SWAGGER_PLUGIN_ORDER)
 public class ApiModelBuilder implements ModelBuilderPlugin {
   private final TypeResolver typeResolver;
+  private final ModelProvider modelProvider;
 
   @Autowired
-  public ApiModelBuilder(TypeResolver typeResolver) {
+  public ApiModelBuilder(TypeResolver typeResolver,
+                         @Qualifier("cachedModels") ModelProvider modelProvider) {
```
dilipkrish: Ideally we don't want plugins to depend on other plugins. While this works this has the possibility of introducing circular dependencies

springfox-swagger-common/src/main/java/springfox/documentation/swagger/schema/ApiModelBuilder.java
```diff
@@ -48,6 +56,19 @@ public void apply(ModelContext context) {
...
+        context.getBuilder().discriminator(annotation.discriminator());
+      }
+
+      // default parent is Void
+      if (!annotation.parent().isAssignableFrom(Void.class)) {
+        Optional<Model> parentModel = modelProvider.modelFor(ModelContext.fromParent(context, this.typeResolver.resolve(annotation.parent())));
+
+        // parent model exist because @ApiModel not required on model
+        // if(parentModel.isPresent()) {
+          context.getBuilder().parent(parentModel.get());
```
dilipkrish: Like I said above this could just be a `ModelRef` and eliminate the need to resolve the model in like 66

rcruzper:
it would be possible to add this feature in the milestone 2.4.0?

dilipkrish:
I could try be if we finish out this pr based on feedback it might be possible

roni-frantchi:
Thanks for this.
We really need this to get our project going. This comes into play especially when you have mixings of accessor types where a parent class is serialized by field access and the child by property - flattening of the properties yields wrong results.

I see this feature is now pushed to milestone 2.5.0, is there an ETA for 2.5.0 ?

Thanks.

roni-frantchi:
@dilipkrish  @Fyro-Ing  I think there's may be an issue with the above - 
After having merged to 2.4.0 and testing the code above, it fails when attempting to inherit form a model not explicitly referenced by any other model explicitly used in the API.
I attempted using ```.additionalModels``` on my ```docket``` but that didn't help.
The only way around it was to actually include a reference to the parent class as a member.

So for the following:
```
@ApiModel(parent = Parent.class)
public class Child extends Parent {
...
}
```

will only work for me if ```Parent``` is explicitly referenced somewhere in the exposed API.
to workaround it I would have to:
```
@ApiModel(parent = Parent.class)
public class Child extends Parent {
    private Parent dummyReference;
...
}
```

Here's the stack trace the above fails if not using the workaround:
```
java.lang.IllegalStateException: Optional.get() cannot be called on an absent value
	at com.google.common.base.Absent.get(Absent.java:47) ~[guava-18.0.jar:na]
	at springfox.documentation.swagger.schema.ApiModelBuilder.apply(ApiModelBuilder.java:69) ~[springfox-swagger-common-2.4.0-poly-2.4.0-gdba3710-2.jar:2.4.0-SNAPSHOT]
	at springfox.documentation.schema.plugins.SchemaPluginsManager.model(SchemaPluginsManager.java:58) ~[springfox-schema-2.4.0-poly-2.4.0-gdba3710-2.jar:2.4.0-SNAPSHOT]
	at springfox.documentation.schema.DefaultModelProvider.modelBuilder(DefaultModelProvider.java:107) ~[springfox-schema-2.4.0-poly-2.4.0-gdba3710-2.jar:2.4.0-SNAPSHOT]
	at springfox.documentation.schema.DefaultModelProvider.modelFor(DefaultModelProvider.java:90) ~[springfox-schema-2.4.0-poly-2.4.0-gdba3710-2.jar:2.4.0-SNAPSHOT]
	at springfox.documentation.schema.DefaultModelProvider.dependencies(DefaultModelProvider.java:115) ~[springfox-schema-2.4.0-poly-2.4.0-gdba3710-2.jar:2.4.0-SNAPSHOT]
	at springfox.documentation.schema.CachingModelProvider.dependencies(CachingModelProvider.java:68) ~[springfox-schema-2.4.0-poly-2.4.0-gdba3710-2.jar:2.4.0-SNAPSHOT]
	at springfox.documentation.spring.web.scanners.ApiModelReader.populateDependencies(ApiModelReader.java:133) ~[springfox-spring-web-2.4.0-poly-2.4.0-gdba3710-2.jar:2.4.0-SNAPSHOT]
	at springfox.documentation.spring.web.scanners.ApiModelReader.read(ApiModelReader.java:76) ~[springfox-spring-web-2.4.0-poly-2.4.0-gdba3710-2.jar:2.4.0-SNAPSHOT]
	at springfox.documentation.spring.web.scanners.ApiListingScanner.scan(ApiListingScanner.java:88) ~[springfox-spring-web-2.4.0-poly-2.4.0-gdba3710-2.jar:2.4.0-SNAPSHOT]
	at springfox.documentation.spring.web.scanners.ApiDocumentationScanner.scan(ApiDocumentationScanner.java:69) ~[springfox-spring-web-2.4.0-poly-2.4.0-gdba3710-2.jar:2.4.0-SNAPSHOT]
	at springfox.documentation.spring.web.plugins.DocumentationPluginsBootstrapper.scanDocumentation(DocumentationPluginsBootstrapper.java:105) ~[springfox-spring-web-2.4.0-poly-2.4.0-gdba3710-2.jar:2.4.0-SNAPSHOT]
	at springfox.documentation.spring.web.plugins.DocumentationPluginsBootstrapper.onApplicationEvent(DocumentationPluginsBootstrapper.java:91) ~[springfox-spring-web-2.4.0-poly-2.4.0-gdba3710-2.jar:2.4.0-SNAPSHOT]
	at springfox.documentation.spring.web.plugins.DocumentationPluginsBootstrapper.onApplicationEvent(DocumentationPluginsBootstrapper.java:53) ~[springfox-spring-web-2.4.0-poly-2.4.0-gdba3710-2.jar:2.4.0-SNAPSHOT]
	at org.springframework.context.event.SimpleApplicationEventMulticaster.invokeListener(SimpleApplicationEventMulticaster.java:163) ~[spring-context-4.2.4.RELEASE.jar:4.2.4.RELEASE]
	at org.springframework.context.event.SimpleApplicationEventMulticaster.multicastEvent(SimpleApplicationEventMulticaster.java:136) ~[spring-context-4.2.4.RELEASE.jar:4.2.4.RELEASE]
	at org.springframework.context.support.AbstractApplicationContext.publishEvent(AbstractApplicationContext.java:381) ~[spring-context-4.2.4.RELEASE.jar:4.2.4.RELEASE]
	at org.springframework.context.support.AbstractApplicationContext.publishEvent(AbstractApplicationContext.java:335) ~[spring-context-4.2.4.RELEASE.jar:4.2.4.RELEASE]
	at org.springframework.context.support.AbstractApplicationContext.finishRefresh(AbstractApplicationContext.java:855) ~[spring-context-4.2.4.RELEASE.jar:4.2.4.RELEASE]
	at org.springframework.boot.context.embedded.EmbeddedWebApplicationContext.finishRefresh(EmbeddedWebApplicationContext.java:140) ~[spring-boot-1.3.2.RELEASE.jar:1.3.2.RELEASE]
	at org.springframework.context.support.AbstractApplicationContext.refresh(AbstractApplicationContext.java:541) ~[spring-context-4.2.4.RELEASE.jar:4.2.4.RELEASE]
	at org.springframework.boot.context.embedded.EmbeddedWebApplicationContext.refresh(EmbeddedWebApplicationContext.java:118) ~[spring-boot-1.3.2.RELEASE.jar:1.3.2.RELEASE]
	at org.springframework.boot.SpringApplication.refresh(SpringApplication.java:766) [spring-boot-1.3.2.RELEASE.jar:1.3.2.RELEASE]
	at org.springframework.boot.SpringApplication.createAndRefreshContext(SpringApplication.java:361) [spring-boot-1.3.2.RELEASE.jar:1.3.2.RELEASE]
	at org.springframework.boot.SpringApplication.run(SpringApplication.java:307) [spring-boot-1.3.2.RELEASE.jar:1.3.2.RELEASE]
	at org.springframework.boot.SpringApplication.run(SpringApplication.java:1191) [spring-boot-1.3.2.RELEASE.jar:1.3.2.RELEASE]
	at org.springframework.boot.SpringApplication.run(SpringApplication.java:1180) [spring-boot-1.3.2.RELEASE.jar:1.3.2.RELEASE]
```

martinsefcik:
Any updates on this ?

roni-frantchi:
See fixes I applied to swagger-core:
https://github.com/swagger-api/swagger-core/pull/1741/commits

thought maybe it'll help this project too - it seems to fix the same symptom as described in my previous comment as well.

benfowler:
Hi. Any word on this PR?

dilipkrish:
@benfowler needs some work... actually may need to be re-implemented

DDani:
Hi @dilipkrish,there is any way we can help to get this PR back to 2.7?. It can be really helpful for many of us :)
Thx!

dilipkrish:
In 2.7.1 perhaps... trying to freeze any new features coming in to release 2.7.0

EduardoSolanas:
We are also interested in this thing, right know it is a blocker and we are considering ditching springfox all together

dilipkrish:
@EduardoSolanas Would you consider submitting a PR. I don't have the bandwidth to work on this and this PR isn't in line with the direction OAS is going in nor is it inline with the direction of this library.

In a side note, IMHO, its not really a good idea to expose DTO's with object hierarchies. It may be convenient on the implementation side.

DDani:
@dilipkrish Perhaps is true, but it simplifies a lot the readability of the code. In the tipical AbstractPet, and their subclassed Dog & Cat it can be a little mess to have a PetDTO at code level, aggregating all the fields when it can be calculated on the fly.

On the other hand I'm agree: let's fix / redo the PR

Fyro-Ing:
if i've time with my job, i'll make changes

roelensw:
Also eagerly waiting for this.  We are currently migrating SOAP interfaces towards REST and need polymorphism for some DTOs.

dilipkrish:
Thanks @Fyro-Ing 🙇 ! Would you mind creating a creating a rough design approach in your PR, that way we can make sure its in line with the librarys future direction. Also it helps with not having lingering PR's that Im hesitant to merge.

dilipkrish:
@DDani 

> but it simplifies a lot the readability of the code.

How? In increases it IMO, because now you have to have conditional logic to route to the right implementation, based on if its a dog or cat. Not to mention the gymnastics required to serialize/deserialize the models.

> In the tipical AbstractPet, and their subclassed Dog & Cat it can be a little mess to have a PetDTO at code level, aggregating all the fields when it can be calculated on the fly.

If you have special logic for Cat and Dog, then arguably there should be a separate endpoint. Its easier than having the client call the same endpoint and have the implementation add an if-then-else to route the request. What you're sharing between `Cat` and `Dog` from the AbstractPet can still be used. 

Assuming you still do the inheritance hierarchy.

```java
class Dog extends AbstractPet {}
class Cat extends AbstractPet {}
```
Then prefer

```java 
void updateDog(Dog dog);
void updateCat(Cat cat);
```

over

```java 
void updatePet(AbstractPet pet);
```
 If you do this, how you've modeled Dog and Cat is moot from a service specification standpoint. You're clients don't have to care about your implementation details i.e. polymorphism.

EduardoSolanas:
In our case, is to provide information about the different types of model the collection contains, for example:
A collection of line items which:
- SaleLineItem contains => id, description, totalPriceIncVat, TotalPriceExVat
- DepositLineItem contains => id, description, totalAmount

mathieuales:
Hi !
What about supporting polymorphism in [OpenAPI 1.2](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/1.2.md#527-model-object) using `subTypes` and `discriminator `?

dilipkrish:
@mathieuales will be happening with OAS 3.0 support

cemo:
@dilipkrish when will OAS 3.0 support be released?

dilipkrish:
@cemo Id say sometime end Sept beg Oct.

roberterdin:
> Assuming you still do the inheritance hierarchy.
> ```
> class Dog extends AbstractPet {}
> class Cat extends AbstractPet {}
> ```
> Then prefer
> ```
> void updateDog(Dog dog);
> void updateCat(Cat cat);
> over
>```

Assume you have an endpoint like `POST /shelter` that takes an array of `AbstractPet`. You can then post all pets you need to in one request which is desirable from a users perspective. If we're speaking about hundreds of different implementations of `AbstractPet` here it would be downright madness to ask the user to use a separate endpoint for each type of pet.

What's the state of this PR? We're evaluating the use of springfox but decent support for polymorphism is a KO criteria. If an implementation is sketched out I can see if I can get somebody to implement this.

elw00d:
Any progress on this ?

dilipkrish:
Working on it

arnokoehler:
My colleague has tried to fix this pull request in https://github.com/rabobank-nederland/springfox this fork. On the Polymorphism branch, but when I try to implement it in a vanilla Spring Boot project, I notice that the request still gets parameters of the parent. Only the models are shown as parent child, but shouldn't the request contain a #ref now to the models? I have World @ApiModel(discriminator = "name", subTypes = {Mars.class, MyWorld.class}) as parent and Mars / MyPlanet as @ApiModel(parent = World.class) childs - @ApiParam(type = "World") on the @RequestBody like:
`World berekenConsumptieveLast(@ApiParam(type = "World") @RequestBody World world);`
and my config has 
`build().additionalModels(typeResolver.resolve(Mars.class), typeResolver.resolve(MyWorld.class));`
But the produced json is:
`{
swagger: "2.0",
info: {
description: "Api Documentation",
version: "1.0",
title: "Api Documentation",
termsOfService: "urn:tos",
contact: { },
license: {
name: "Apache 2.0",
url: "http://www.apache.org/licenses/LICENSE-2.0"
}
},
host: "localhost:8080",
basePath: "/",
tags: [
{
name: "basic-error-controller",
description: "Basic Error Controller"
},
{
name: "berekenen planeet",
description: "	"
}
],
paths: {
/error: {
get: {
tags: [
"basic-error-controller"
],
summary: "errorHtml",
operationId: "errorHtmlUsingGET",
produces: [
"text/html"
],
responses: {
200: {
description: "OK",
schema: {
$ref: "#/definitions/ModelAndView"
}
},
401: {
description: "Unauthorized"
},
403: {
description: "Forbidden"
},
404: {
description: "Not Found"
}
}
},
head: {
tags: [
"basic-error-controller"
],
summary: "errorHtml",
operationId: "errorHtmlUsingHEAD",
consumes: [
"application/json"
],
produces: [
"text/html"
],
responses: {
200: {
description: "OK",
schema: {
$ref: "#/definitions/ModelAndView"
}
},
204: {
description: "No Content"
},
401: {
description: "Unauthorized"
},
403: {
description: "Forbidden"
}
}
},
post: {
tags: [
"basic-error-controller"
],
summary: "errorHtml",
operationId: "errorHtmlUsingPOST",
consumes: [
"application/json"
],
produces: [
"text/html"
],
responses: {
200: {
description: "OK",
schema: {
$ref: "#/definitions/ModelAndView"
}
},
201: {
description: "Created"
},
401: {
description: "Unauthorized"
},
403: {
description: "Forbidden"
},
404: {
description: "Not Found"
}
}
},
put: {
tags: [
"basic-error-controller"
],
summary: "errorHtml",
operationId: "errorHtmlUsingPUT",
consumes: [
"application/json"
],
produces: [
"text/html"
],
responses: {
200: {
description: "OK",
schema: {
$ref: "#/definitions/ModelAndView"
}
},
201: {
description: "Created"
},
401: {
description: "Unauthorized"
},
403: {
description: "Forbidden"
},
404: {
description: "Not Found"
}
}
},
delete: {
tags: [
"basic-error-controller"
],
summary: "errorHtml",
operationId: "errorHtmlUsingDELETE",
produces: [
"text/html"
],
responses: {
200: {
description: "OK",
schema: {
$ref: "#/definitions/ModelAndView"
}
},
204: {
description: "No Content"
},
401: {
description: "Unauthorized"
},
403: {
description: "Forbidden"
}
}
},
options: {
tags: [
"basic-error-controller"
],
summary: "errorHtml",
operationId: "errorHtmlUsingOPTIONS",
consumes: [
"application/json"
],
produces: [
"text/html"
],
responses: {
200: {
description: "OK",
schema: {
$ref: "#/definitions/ModelAndView"
}
},
204: {
description: "No Content"
},
401: {
description: "Unauthorized"
},
403: {
description: "Forbidden"
}
}
},
patch: {
tags: [
"basic-error-controller"
],
summary: "errorHtml",
operationId: "errorHtmlUsingPATCH",
consumes: [
"application/json"
],
produces: [
"text/html"
],
responses: {
200: {
description: "OK",
schema: {
$ref: "#/definitions/ModelAndView"
}
},
204: {
description: "No Content"
},
401: {
description: "Unauthorized"
},
403: {
description: "Forbidden"
}
}
}
},
/world/search/: {
post: {
tags: [
"berekenen planeet"
],
summary: "Met een naam krijg je en planeet",
operationId: "berekenConsumptieveLastUsingPOST",
consumes: [
"application/json"
],
produces: [
"application/json"
],
parameters: [
{
name: "name",
in: "query",
required: true,
type: "string"
},
{
name: "size",
in: "query",
required: false,
type: "integer",
format: "int32"
}
],
responses: {
200: {
description: "Hello planet",
schema: {
$ref: "#/definitions/World"
}
},
201: {
description: "Created"
},
400: {
description: "Welke planeet?"
},
401: {
description: "Unauthorized"
},
403: {
description: "Verboden planeet!"
},
404: {
description: "Not Found"
},
500: {
description: "Oeps blackhole sun"
}
}
}
}
},
definitions: {
Mars: {
title: "Mars",
allOf: [
{
$ref: "#/definitions/World"
},
{
type: "object",
required: [
"color",
"deatlyWeather",
"name"
],
properties: {
color: {
type: "string",
description: "De kleur rood",
allowEmptyValue: false
},
deatlyWeather: {
type: "boolean",
example: false,
description: "acid storm?",
allowEmptyValue: false
},
name: {
type: "string",
description: "De naam",
allowEmptyValue: false
},
size: {
type: "integer",
format: "int32",
description: "De afmeting",
allowEmptyValue: false
}
},
title: "Mars"
}
]
},
ModelAndView: {
type: "object",
properties: {
empty: {
type: "boolean"
},
model: {
type: "object"
},
modelMap: {
type: "object",
additionalProperties: {
type: "object"
}
},
reference: {
type: "boolean"
},
status: {
type: "string",
enum: [
"100",
"101",
"102",
"103",
"200",
"201",
"202",
"203",
"204",
"205",
"206",
"207",
"208",
"226",
"300",
"301",
"302",
"303",
"304",
"305",
"307",
"308",
"400",
"401",
"402",
"403",
"404",
"405",
"406",
"407",
"408",
"409",
"410",
"411",
"412",
"413",
"414",
"415",
"416",
"417",
"418",
"419",
"420",
"421",
"422",
"423",
"424",
"426",
"428",
"429",
"431",
"451",
"500",
"501",
"502",
"503",
"504",
"505",
"506",
"507",
"508",
"509",
"510",
"511"
]
},
view: {
$ref: "#/definitions/View"
},
viewName: {
type: "string"
}
},
title: "ModelAndView"
},
MyWorld: {
title: "MyWorld",
allOf: [
{
$ref: "#/definitions/World"
},
{
type: "object",
required: [
"name",
"normal"
],
properties: {
name: {
type: "string",
description: "De naam",
allowEmptyValue: false
},
normal: {
type: "boolean",
example: false,
description: "is my world loco?",
allowEmptyValue: false
},
size: {
type: "integer",
format: "int32",
description: "De afmeting",
allowEmptyValue: false
}
},
title: "MyWorld"
}
]
},
View: {
type: "object",
properties: {
contentType: {
type: "string"
}
},
title: "View"
},
World: {
type: "object",
required: [
"name"
],
discriminator: "name",
properties: {
name: {
type: "string",
description: "De naam",
allowEmptyValue: false
},
size: {
type: "integer",
format: "int32",
description: "De afmeting",
allowEmptyValue: false
}
},
title: "World"
}
}
}`

marcelstoer:
As this PR does not reference a particular issue it's not entirely clear to me _which_ problems it tries to solve. I'm positive, though, that we're currently stuck with the same problem as the ones the PR tries to address.

We've got a Swagger definition like so:

```yaml
definitions:
  Message:
    type: object
    discriminator: transferableType
    properties:
      transferableType:
    ...
  MyMessage:
    allOf:
      - $ref: '#/definitions/Message'
   ....
```

And the respective (Spring Boot) Java model like so:

```java
@ApiModel(subTypes = {MyMessage.class}, discriminator = "transferableType")
public class Message {
...
@ApiModel(parent = Message.class)
public class MyMessage extends Message {
```

Yet, when Springfox generates the API docs **there's no definition for `Message` and its properties are merged into `MyMessage`**.

That's this issue, right?

**Context**
When Springfox is used purely for documentation purposes it may be negligible that definition hierarchies are squashed. For us it's disastrous, though. With the help of [assertj-swagger](https://github.com/RobWin/assertj-swagger) we're validating whether an implementation is compliant to the service contract. For that we compare the design-first Swagger model from YAML with the actual Swagger model represented by the JSON generated by Springfox. This will always fail as the "implementation" doesn't seem to provide a `Message` model. As a workaround I looked at [`additionalModels` in Docket](https://github.com/springfox/springfox/blob/master/springfox-spring-web/src/main/java/springfox/documentation/spring/web/plugins/Docket.java#L413) but I don't how that is supposed to be used.

cemo:
@dilipkrish I know you are not actively working on this project butthis PR is really important and will solve a bunch of issues on are side too. Would you please review it?

dilipkrish:
@cemo I am actively working on this project. Just not this PR :) This solution will not work when supporting OpenAPI 3.0 thats why its not pulled in.

ralfhecktor:
@dilipkrish Why will it not work? Where do you see any problems? Can you please elaborate on that a bit. Whats with backwards compatibility for OAS 2?

dilipkrish:
@ralfhecktor composition based inheritance was introduced in 3.0 (allOf, anyOf etc.) with better schema support. Discriminator based inheritance isn't ideal when there is better support. I'd rather think the problem through "whole brain" than hack the solution together.

cemo:
@dilipkrish do you have any idea when 3.0 support will be landed?

dilipkrish:
But Im hoping in June. If I get some help with PR's etc. may be earlier

raderio:
> If I get some help with PR's etc. may be earlier

How can we help?

raderio:
>Discriminator based inheritance isn't ideal when there is better support.

It is still supported in OpenAPI 3 https://swagger.io/docs/specification/data-models/inheritance-and-polymorphism/ Discriminator section, if it is better than why they support it?

Also discriminator is supported by Jackson https://gist.github.com/christophercurrie/8939489

dilipkrish:
Im not denying there is support, just suggesting there are better ways of doing it @raderio

raderio:
If there are better ways of doing it, why OpenAPI 3 supports it, why they just don't dropped it? Maybe because in some cases discriminator fits better?
I think library/specification should give tools to allow to implement different solutions, to use or not to use the discriminator approach is choice of application side.

dilipkrish:
@Fyro-Ing thanks for your PR, while I didnt incorporate all of your changes I cherry picked the important changes you contributed to the `ModelMapper` 🤘

raderio:
@dilipkrish coooool 
It will be available in next release? an ETA for next release?
Is it possible somehow to test/use before release?

dilipkrish:
@raderio targetting this weekend. Yes feedback will be awesome! You can [use the SNAPSHOT build](http://springfox.github.io/springfox/docs/current/#dependencies)

raderio:
@dilipkrish I've tested, it works, big thanks.

cemo:
@raderio If you don't mind, would you please share your use case?

dilipkrish:
Thanks @raderio for confirming! 👍 

@cemo here is the example use case  👇 

https://github.com/springfox/springfox/commit/eb6ca4dfbea43a03a1dfb1cd58ef35cfc0af9792

diogosouza:
Guys, is there any idea if this would by applied to springfox-swagger-ui as well? 
Testing here, it doesn't change the ui's look and feel...

dilipkrish:
@diogosouza What are you seeing/not-seeing?

MrDefinite:
Hi @dilipkrish, 
I can see the the child class model in swagger ui, which has all the properties inherited from parent class. But there is no description telling which parent it is inherited from the ui. Is it possible to have such a tag on ui?

dilipkrish:
@MrDefinite not if its not supported in the spec

zpf7879:
Hi @dilipkrish,
I have the same question as @MrDefinite. In this situation, it is difficult for my customer to use the API esp. for the PUT operation, e.g. create(Animal). First, the customer needs to find the derived class, and he also needs to remember to provide the /@ discriminator.

{
  "@type": "Cat",
  "id": 2,
  "color": null,
  "parent": null,
}

Any suggestion?

dilipkrish:
No. Don't use inheritance in an API. It exposes the type system and implementation language internals.

Issue #356, opened 2014-07-08
Support for JSR-303 (Java Bean Validation) ?


omayevskiy:
Hi is there any support for JSR-303 (Java Bean Validation) ?

Given

A controller method
```
import javax.validation.Valid;
@RequestMapping(method = RequestMethod.PUT)
    public void create( @RequestBody @Valid MyDTO myDTO) {
        // do something with myDTO
    }
```

A command object
```
import javax.validation.constraints.NotNull;
public class MyDTO {

    @NotNull
    private Long myId;

    @NotNull
    private String myValue;
```

Swagger produces
```
MyDTO {
   myId (integer, optional),
   myValue (string, optional)
}
```

But should be

```
MyDTO {
   myId (long?, required),
   myValue (string, required)
}
```

I could add swagger annotations to MyDTO's fields, but then the NotNull/required logic would be duplicated and potentially leads to documentation inconsistence.

Thx for reading ;-)

jasonchaffee:
It would be nice if it could handle this not only in the model, but for Query, Path, Matrix params as well.  Also, it would nice if it would handle more than @NotNull and could handle the other annotations as part of the spec.  See http://docs.oracle.com/javaee/7/api/javax/validation/constraints/package-summary.html

Then, the clients of the swagger spec, such as the swagger-ui could validate the fields as they are entered without needing to make a call to the server.

adrianbk:
Relates to this issue: https://github.com/springfox/springfox/issues/311

As far as I can tell none of the JSR-303 constraint's are supported by springfox (including @NotNull).
It should be relatively straight forward to offer rudimentary support for these constraints but we would have to provide a way to disable the feature because there are scenarios, such as using [groups](https://docs.jboss.org/hibernate/validator/4.0.1/reference/en/html_single/#validator-usingvalidator-validationgroups) to conditionally apply validation rules based on runtime conditions, where it is not feasible to try to infer validation constraints.

@jasonchaffee would you be interesting in contributing to your idea?

dilipkrish:
I prefer this to be separate module ```springfox-validations```. Should be simple enough to use the plugin models to enrich the service descriptions. 

Don't want it to be a concern in the core modules, just because it can be released incrementally and possibly out-of-band.

jasonchaffee:
I agree that it makes sense to be in a separate module.

jasonchaffee:
[...]

PR #1159, opened 2016-01-30
Add basic Bean validation api (JSR-303) support 

jfiala:
see https://github.com/springfox/springfox/issues/356

Added support for 
* @NotNull
* @Min
* @Max
* @Size

### DIFF
--- a/springfox-schema/src/main/java/springfox/documentation/schema/property/OptimizedModelPropertiesProvider.java
+++ b/springfox-schema/src/main/java/springfox/documentation/schema/property/OptimizedModelPropertiesProvider.java
@@ -38,6 +38,7 @@ import com.google.common.annotations.VisibleForTesting;
 import com.google.common.base.Function;
 import com.google.common.base.Optional;
 import com.google.common.base.Predicate;
+import com.google.common.collect.FluentIterable;
 import org.slf4j.Logger;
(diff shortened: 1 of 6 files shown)


dilipkrish:
Thanks @jfiala I polished it up some and pushed your pr via https://github.com/springfox/springfox/commit/727e4256b426ad458efbac8907f8a1d05b338cfc

dilipkrish:
Would you please verify I've got it right

jfiala:
It works greatly, but only after adding two null-checks in BeanValidators.java:

``` java
if (propertyDefinition.isPresent()) {
        // 1. not-null-check
    	BeanPropertyDefinition def = propertyDefinition.get();
    	
    	if (def!=null) {
               // 2. not-null-check
    		AnnotatedField field = def.getField();
    		if (field!=null) {
    			notNull = FluentIterable
      		          .from(propertyDefinition.get().getGetter().annotations())
      		          .filter(annotationType)
      		          .first()
      		          .or(FluentIterable
      		              .from(field.annotations())
      		              .filter(annotationType)
      		              .first());
    		}
    		
    	}
      
    }

```

dilipkrish:
Thanks for catching that... fixed via 281e70f27e6584372ec274019bb4ad6ae6b05829

PR #1163, opened 2016-02-03
Updated the Docket XML Configuration Documentation

kellydavid:
The sample Springfox configuration class uses the `@Configuration` annotation. This requires the xml configuration to have an extra element so that Spring can carry out post processing on the configuration class for Springfox. According to the [Spring documentation](http://docs.spring.io/autorepo/docs/spring/current/javadoc-api/org/springframework/context/annotation/Configuration.html): 

> In the example above, <context:annotation-config/> is required in order to enable ConfigurationClassPostProcessor and other annotation-related post processors that facilitate handling `@Configuration` classes.


dilipkrish:
Thanks @kellydavid

PR #1165, opened 2016-02-04
ParameterDefaultReader in swagger does not honor other annotations usage of DefaultValue

ctruzzi:
When annotations other then ApiParam are used, the defaultValue set by them is not honored. 
If an annotation such as Springs @RequestParam is used with a defaultValue set and @APIParam is not used, the current implementation will override the defaultValue of the parameterBuilder to  empty string. 

Note: springfox.documentation.spring.web.readers.parameter.ParameterDefaultReader correctly evaluates the defaultValue of RequestParam and correctly assigns the defaultValue (not overriding previous values set if the annotation doesn't exist).

### DIFF
--- a/springfox-swagger-common/src/main/java/springfox/documentation/swagger/readers/parameter/ParameterDefaultReader.java
+++ b/springfox-swagger-common/src/main/java/springfox/documentation/swagger/readers/parameter/ParameterDefaultReader.java
@@ -43,7 +43,7 @@ public class ParameterDefaultReader implements ParameterBuilderPlugin {
     String defaultValue = findAnnotatedDefaultValue(methodParameter);
     boolean isSkip = ValueConstants.DEFAULT_NONE.equals(defaultValue);
     if (!isSkip) {
-      context.parameterBuilder().defaultValue(nullToEmpty(defaultValue));
+      context.parameterBuilder().defaultValue(defaultValue);
     }
   }
 


ctruzzi:
hi @dilipkrish,
I am unsure if those tests were not working properly, but I was able to make a modification to ParameterBuilder#defaultValue that works with those 1.2 annotation tests.

dilipkrish:
@ctruzzi I thought about doing that fix that you did, but then what if we had the case where we want the default value to be a blank string?

ctruzzi:
I believe it would work properly as long as you do not have a conflicting annotation, since the initial value is set to null, as long as that value does not get set to anything prior to defaultValue = "" it would assign to that properly in this conditional argument. 

However if we had something like
@RequestParam(defaultValue="value")
@APIParam(defaultValue="")

The new code revision would keep defaultValue assigned to value. I am fine with either code modification, and the first one makes more sense if indeed those tests were doing incorrect checks.

dilipkrish:
Philosophically the way it works is that we layer information on top of the other. The layer happens based on priority.
1. Spring Web annotations are applied first
2. Bean Validation annotations go next (recently introduced)
3. Swagger annotations are applied last

So the way I see it playing out is that if someone puts in the swagger annotation that overwrites the spring default value, then it will be apparent where the problem is. They would need to set the `@ApiParam#defaultValue` to a non-default value. Let me know if that makes sense.

This way there is no conflicting annotations. Its always predictable and there is not conditional code to handle just this particular edge case. Hope that makes sense

ctruzzi:
Since there is that defined priority, I would agree that the initial pull request would be the most correct one to keep the code simple and working properly. 

Is there an ETA for 2.4 milestone? (Currently using 2.0.1 and require some changes from 2.0.2 but want a single upgrade time).

dilipkrish:
Its set for  March 1st, but I want to get as many features as possible by then.

Issue #1210, opened 2016-03-08
Context-Path missing when behind Proxy


matonthecat:
Hey,

I'm using Springfox Swagger2 v2.4.0 with a Zuul as the reverse proxy. My aim is to implement a RESTService documented by swagger. 

Direct links on the host look like this:
`http://service-host:8080/path`

whereas links to be routed via Zuul look like this:
`http://zuul-host:8080/service-context/path`

While the swagger-ui generates the correct URLs when called from the service-host directly. However, when called via the reverse proxy, it fails to produce correct try-out links. The links produced do contain the right server/port but not the routes as assigned on the reverse proxy (albeit the `X-Forwarded-*` are all set) like this:

`http://zuul-host:8080/path`

This is due to the implementation of the  [Swagger2Controller](https://github.com/springfox/springfox/blob/master/springfox-swagger2/src/main/java/springfox/documentation/swagger2/web/Swagger2Controller.java#L88) where the context path is ignored for some reason.

```java
UriComponents uri = componentsFrom(servletRequest);
String host = uri.getHost();
int port = uri.getPort();
if (port > -1) {
    return String.format("%s:%d", host, port);
}
```

As I saw in debugging mode, the generated `UriComponents` in fact do contain the required path altogether. I would appreciate, if that path would be the one returned, or someone implement context awareness in some other way.

Thx,

mat

dilipkrish:
@matonthecat I haven't had time to try it out myself or create a sample but there has been much success based on using the approach that the jhipster folks have taken. [Check this issue](https://github.com/springfox/springfox/issues/1204#issuecomment-193275322) for more details.

matonthecat:
Hi @dilipkrish ,

thank you for your swift response. 

I found that part of the solution will be configuring [a centralized swagger-resources as in issue 2990 of jhipster](https://github.com/jhipster/generator-jhipster/issues/2990). Then we could have some central swagger-ui pointing to this resource. 

Nevertheless, configured this way the issue stated above remains: all links generated by the swagger-ui will not contain the context path on the reverse proxy.

This could be fixed easiest by setting the `basePath` in the JSON returned by Swagger2Controller.

What do you think?

dilipkrish:
@matonthecat Not really sure what the problem is. 

- The basePath is [overridable by providing a custom  path provider](http://springfox.github.io/springfox/javadoc/current/springfox/documentation/spring/web/plugins/Docket.html#pathProvider(springfox.documentation.PathProvider)). 
[...]

PR #1217, opened 2016-03-15
adds basePath from x-forwarded-prefix for reverse proxy scenarios

matonthecat:
Hi, 

this is my suggested solution for the [issue 1210](https://github.com/springfox/springfox/issues/1210). 

It fixes generated URLs in the swagger-ui when called from behind a reverse proxy (Netflix Zuul).

Hope for addition...
mat


dilipkrish:
Thank you for the PR _AND_ for being thorough! :+1:

matonthecat:
awesome, many thanks for merging :thumbsup:

sonata82:
Can this simple, elegant and effective solution be added to the documentation? Or is this problem not considered to be part of Springfox?

PR #1218, opened 2016-03-15
Only encode the API key when passed in URL query

franklloydteh:
I am current passing tokens through the header. Everything is working using the current app but when trying it on Swagger UI, I noticed that the tokens changed.

### DIFF
--- a/springfox-swagger-ui/src/web/js/springfox.js
+++ b/springfox-swagger-ui/src/web/js/springfox.js
@@ -49,7 +49,7 @@ $(function() {
     initializeBaseUrl();
 
     function addApiKeyAuthorization() {
-      var key = encodeURIComponent($('#input_apiKey')[0].value);
+      var key = window.apiKeyVehicle == 'query ? encodeURIComponent($('#input_apiKey')[0].value) : $('#input_apiKey')[0].value;
       if (key && key.trim() != "") {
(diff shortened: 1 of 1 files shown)


dilipkrish:
@franklloydteh Thanks for the PR!! :+1: Not sure why the syntax highlighting is off. Have you tested that this works?

franklloydteh:
The syntax highlighting was off because there was a missing single quote. Sorry I did not clone the project and just manually edited the file in github. I am not sure on how this should be tested. I just manually edited the source in chrome :1234:

dilipkrish:
My concern was mainly the syntax highlighting. Any who, would you also elaborate how you encountered this issue? I ask because in general API keys are ascii characters and should not be affected by url encoding. i.e it shouldn't have spaces `/` etc.

dtritus:
I've encountered this issue when tried to send API key as `Authorization` header like `Bearer <token>`.
As you can see there is space between `Bearer` and `<token>` and it's being encoded as `%20` right now.

dilipkrish:
@dtritus thanks!

BryceMehring:
@dilipkrish, When is this change going to be included in a release? I'm really needing this functionality.

dilipkrish:
I'm trying to wrap up the release in the next week but it really depends on the outstanding bugs.

Issue #1264, opened 2016-04-14
Swagger2 json file contains empty license object when ApiInfo has no license info


wgreven-ibr:
Hi,

When you have an ApiInfo object without any license information the resulting swagger2 definition will contain an empty license object. That makes the swagger json invalid.

I think it's better not to add the license object in this case. I will submit a pull request for this change.

Best regards,
Wilco

PR #1265, opened 2016-04-14
Don't include a license object when both license and licenseUrl are e…

wgreven-ibr:
Implements the proposed fix in issue #1264.


dilipkrish:
@wgreven-ibr thank you for the PR! :metal: for the included tests!

Issue #1296, opened 2016-05-02
Support for JSR-303: @Pattern annotation


ashutosh-shirole:
For starters, this is a great library and has been of great help.
I am currently using 2.4.0 version and springfox-bean-validators plugin for JSR 303 bean validation annotations @NotNull @Min @Max.

Have seen [#356](https://github.com/springfox/springfox/issues/356) and comments on [this](https://github.com/springfox/springfox/commit/04065109fb8ce566e7fa8016a00a31202e733d37) commit, which states the support for @Pattern annotation is not available. 

I could not find ModelPropertyBuilderPlugin for Pattern annotation. Am I missing out on something or this annotation is not yet supported?

dilipkrish:
Its not supported.... but it could be :) PR?

jjzabkar:
Perhaps this issue should be tagged as "can-use-for-docs"?
https://github.com/springfox/springfox/labels/can-use-for-docs

Related: https://github.com/springfox/springfox/issues/1227

igor-sokolov:
> Its not supported.... but it could be :) PR?
Here you are! (:

Issue #1231, opened 2016-03-21
JSR-303 support: @NotNull ignored when @ApiModelProperty is present


jfiala:
Version: 2.4.0

1.) @NotNull alone works:
@NotNull
private Date createTimestamp;

2.) @NotNull + @ApiModelProperty without required-Attribute:
@NotNull
@ApiModelProperty(value="Timestamp of creation")
private Date createTimestamp;

Now @NotNull is ignored, although the required-attribute is not set in @ApiModelProperty:
Here @NotNull should be picked up, because there is no indication of the required-attribute value.

3.) Invald combination: @NotNull + @ApiModelProperty with required-Attribute set to false
This combination is in fact invalid, maybe a warning could be logged that the not-null indicators are not consistent:
@NotNull
@ApiModelProperty(value="Timestamp of creation", required=false)
private Date createTimestamp;

dilipkrish:
@jfiala this is working as designed. The plugins work in a particular order and layer information on top of each other.
- First the spring annotations have their say @RequestMapping, @RequestParam etc.
- Then the JSR-303 plugins have their say @NotNull etc.
- Lastly the swagger annotations that have the most specific information like description etc. 

These are enforced by the plugin ordering and priority. Also in this case the default value for the `@ApiModelProperty#required` is [`false`](https://github.com/swagger-api/swagger-core/blob/master/modules/swagger-annotations/src/main/java/io/swagger/annotations/ApiModelProperty.java#L80). Since they don't have a tri-state annotation value that can represent a not-specified value, there is no way for springfox to make that determination. That is why you see the behavior that you're seeing.

jfiala:
I understand that this is the current behaviour.
However, as I would like to use JSR-303 only, it would be great to allow this. Currently I have to also add the required-attribute if  I add documentation, enforcing redundant input.

My suggestions are:
a) (Optionally) put the JSR-303 plugins at the end so they can overrule the others

b) The swagger-annotations should query if something is already determined as required. If this is the case, it should not mark it as non-required.
Because if something has already been marked as required, why should the Swagger annotations be allowed to overrule this?

dilipkrish:
@jfiala This can be a little tricky. The reasons being sometimes we want `allowable values` from swagger annotations, sometimes we may want the range specified via a JSR-303 annotation. Don't know if its easy to make the ordering configurable. Let me know if you have any suggestions. 

The reason we picked swagger annotations last is that we want to use that as a case for overlaying mostly description information. 

[...]

PR #1314, opened 2016-05-16
Support for JSR-303: @Pattern annotation #1296

igor-sokolov:
#### What's this PR do/fix?
It introduces a basic support of `@Pattern` annotation. It was discussed in the issue: https://github.com/springfox/springfox/issues/1296
Yes,  added unit tests similar to already existed for `SizeAnnotationPlugin`. Additionally I tested it locally in project that uses Springfox for document generation. The result YAML file contained  pattern ("pattern":"^[A-Z]{3}$"}). It was also rendered on UI but it was not the perfect, I guess due to overlapping in hint sections, one from first field (double) and the second from message field with pattern restriction. 
I didn't performed any validation of provided in `@Pattern` annotation value, the value just extracted and transferred to Swagger Model object. Ideally it should be validated and also checked that this is not simple a valid RegExp but valid as well for JavaScript.  
https://github.com/springfox/springfox/issues/1296

### DIFF
--- a/springfox-bean-validators/src/main/java/springfox/bean/validators/configuration/BeanValidatorPluginsConfiguration.java
+++ b/springfox-bean-validators/src/main/java/springfox/bean/validators/configuration/BeanValidatorPluginsConfiguration.java
@@ -22,6 +22,7 @@ import org.springframework.context.annotation.Bean;
 import org.springframework.context.annotation.Configuration;
 import springfox.bean.validators.plugins.MinMaxAnnotationPlugin;
 import springfox.bean.validators.plugins.NotNullAnnotationPlugin;
+import springfox.bean.validators.plugins.PatternAnnotationPlugin;
 import springfox.bean.validators.plugins.SizeAnnotationPlugin;
 
 @Configuration
@@ -39,5 +40,10 @@ public class BeanValidatorPluginsConfiguration {
   @Bean
   public NotNullAnnotationPlugin notNullPlugin() {
     return new NotNullAnnotationPlugin();
+  }
+
+  @Bean
+  public PatternAnnotationPlugin patternPlugin() {
+    return new PatternAnnotationPlugin();
   }
 }
--- a/springfox-bean-validators/src/main/java/springfox/bean/validators/plugins/PatternAnnotationPlugin.java
+++ b/springfox-bean-validators/src/main/java/springfox/bean/validators/plugins/PatternAnnotationPlugin.java
@@ -1,0 +1,71 @@
+/*
+ *
+ *  Copyright 2015 the original author or authors.
+ *
+ *  Licensed under the Apache License, Version 2.0 (the "License");
+ *  you may not use this file except in compliance with the License.
+ *  You may obtain a copy of the License at
+ *
+ *         http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *  Unless required by applicable law or agreed to in writing, software
+ *  distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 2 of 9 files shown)


igor-sokolov:
Guys, can you please help me with this PR. When I launched full build on local machine I've got the same error in test cases and it was in the method:
```java
  def "Model substitution registers new rules"() {
    when:
      new Docket(DocumentationType.SWAGGER_12)
              ."${method}"(*args)
              .configure(contextBuilder)
    then:
      context().alternateTypeProvider.rules.size() == expectedSize

    where:
      method                    | args                               | expectedSize
      'genericModelSubstitutes' | [ResponseEntity.class, List.class] | 9
      'directModelSubstitute'   | [LocalDate.class, Date.class]      | 8
  }
``` 

Originally the numbers there were 'genericModelSubstitutes' - 8 and 'directModelSubstitute' - 7. Not sure what it does mean but according to stack tracing the real values were 9 and 7 accordingly. I made decision for myself that this is number of plugins triggered and because I've added one more plugin the numbers should be increased. It worked on my local machine, but I see it doesn't work on CI environment.

What am I doing wrong?

dilipkrish:
Yeah need to fix this test to be jdk agnostic. Right now only jdk 7 is supported.

dilipkrish:
Thanks for the PR :metal:

igor-sokolov:
> Yeah need to fix this test to be jdk agnostic. Right now only jdk 7 is supported.

Hmm, Not sure I know how to fix it. I compiled locally even with jdk 8. Any suggestions/hints?

ashutosh-shirole:
As far as I have seen the implementation of Pattern, I think this might overwrite if there is @Size annotation along with @Pattern annotation on a field. I am not sure but will have to test.

igor-sokolov:
Hi @ashutosh-shirole, you are right. this is behavior that I decided is the best from targets I see in Springfox. I was guided by an idea that RegExp validation is more powerful than  `@Size`, I mean you can do everything what you can with `@Size` and a little bit more:-) For example instead to write `@Size(min = 1, max = 3)` you could provide RegExp like `@Pattern(regexp = "^.{1,3}$")` or better add more specific information like "only digits are allowed": `@Pattern(regexp = "^\d{1,3}$")`. More complexity adds to this questions a case if there are both annotations assigned to the same method/field and they could be in inconsistent state to each other. So in this case the easiest and still correct way to solve this problem seems to me use only `@Pattern` when both annotations assigned, because Springfox this is not in competence to resolve discrepancies in this kind of situation and the main task is only to document API.  If it happens then likely the validation logic in application should not pass correct values and developer will fix the discrepancy. I hope my process of thinking and decision makes sense to you.

dilipkrish:
@igor-sokolov @ashutosh-shirole seems like both PRs end up doing the same thing. 

My 2 💰 is that having 2 mutually exclusive plugins is always better than competing plugins that can overwrite the other ones values as evidenced by [this discussion](https://github.com/springfox/springfox/issues/1231).

It seems like the right thing to do here is to make `pattern` a first class citizen of `ModelProperty` as it is a first class property on the OAS models as well.

igor-sokolov:
Guys, OK. It sounds reasonable to me, especially because the model in Swagger (OAS) has all three type of fields on `StringProperty`, I mean
- io.swagger.models.properties.StringProperty#_enum
- io.swagger.models.properties.StringProperty#minLength and io.swagger.models.properties.StringProperty#maxLength
- io.swagger.models.properties.StringProperty#pattern

I took a look at PR from @ashutosh-shirole and it seems it does exactly what we discussed here. Probably then it makes sense to switch to that PR and close mine.

PR #1316, opened 2016-05-17
#1296: Adding Pattern annotation support

ashutosh-shirole:
#### What's this PR do/fix?
Provide support for @Pattern annotation of JSR-303. #1296
There are new Unit tests added for Pattern tests, Similar to SizeAnnotation plugin.
Some changes had to be made in core module. ModelPropertyBuilder, ModelProperty from springfox-core and ModelMapper from springfox-swagger2 and their tests had to be changed.
#1296 and #1314

### DIFF
--- a/springfox-bean-validators/src/test/java/springfox/bean/validators/plugins/models/PatternTestModel.java
+++ b/springfox-bean-validators/src/test/java/springfox/bean/validators/plugins/models/PatternTestModel.java
@@ -1,0 +1,47 @@
+/*
+ *
+ *  Copyright 2015 the original author or authors.
+ *
+ *  Licensed under the Apache License, Version 2.0 (the "License");
+ *  you may not use this file except in compliance with the License.
+ *  You may obtain a copy of the License at
+ *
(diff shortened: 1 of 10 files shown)


springfox-bean-validators/src/test/java/springfox/bean/validators/plugins/models/PatternTestModel.java
```diff
@@ -0,0 +1,47 @@
...
+ */
+package springfox.bean.validators.plugins.models;
+
+import javax.validation.constraints.Pattern;
+
+/**
+ * @author : ashutosh 
+ * 29/04/2016
+ */
+public class PatternTestModel {
```
igor-sokolov: Probably it would be good to add here an example when we have both annotations `@Size` and `@Pattern` assigned to one field and check that both information is kept in the context? This might help to understand to somebody who will be interesting how it should work in this specific case. How do you think?

dilipkrish:
Thanks for the pr! More importantly thanks for working together to get this working @ashutosh-shirole and @igor-sokolov! Sorry it took so long.

PR #1341, opened 2016-06-12
Adding @Past/@Future support to Bean Validators

boeboe:
#### What's this PR do/fix?
It adds @Past and @Future Bean Validator support

Yes, unit tests available in _springfox-bean-validators/src/test_

I needed them on my project

No issues, just added functionality which I wanted to share


dilipkrish:
@boeboe Everything looks good with this PR but I had a question. I just realized that the past and future plugins only contribute to, if a given property is required or not. I was expecting to see either an example or a pattern for regex. Is that intentional or is it an oversight? This seems a very subjective change and in fact is very specific to your project.

dilipkrish:
@boeboe ping!

boeboe:
Feel free to drop the PR, I didn't use it as intended firstly.

Br, Bart

dilipkrish:
@boeboe the intent of this PR is still useful! I'm going to leave it open and it would be great if you could flush out the implementation when you find time!

dilipkrish:
Closing as indicated

PR #1350, opened 2016-06-21
Complex type support by @ModelAttribute

MaksimOrlov:
This code adds functionality to unwrap complex parameters, that were annotated with @ModelAttribute annotation.

Example:
```java
Parameter List<MyObject> myObject;
```
Where:

```java
MyObject {
    String a;
    Integer b;
    List<MyObject2> myObject2;
}

MyObject2 {
    Long c;
}
```
Will be unwrapped to:
```
myObject[#ind].a

myObject[#ind].b

myObject[#ind].myObject2[#ind].c
```

### DIFF
--- a/springfox-spring-web/src/main/java/springfox/documentation/spring/web/readers/parameter/ModelAttributeParameterExpander.java
+++ b/springfox-spring-web/src/main/java/springfox/documentation/spring/web/readers/parameter/ModelAttributeParameterExpander.java
@@ -221,7 +221,7 @@ public class ModelAttributeParameterExpander {
     ResolvedType alternativeType = alternateTypeProvider.alternateFor(resolvedType);
     Class<?> erasedType = alternativeType.getErasedType();
     if (type != erasedType) {
-      LOG.debug("Found alternative type [{}] for field: [{}-{}]", erasedType, field);
+      LOG.debug("Found alternative type [{}] for field: [{}-{}]", erasedType, field, type);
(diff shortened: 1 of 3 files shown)


MaksimOrlov:
Hi. I'm sorry for the delay with answer.

At the swagger-ui or some other tools we will get something like:

myObject[#ind].myObject2[#ind].c

#ind - it's **only** my suggestion to mark array's indexes.

I'm not sure, that swagger-us's function "**Try it out!**" will work for requests with parameters like that.

My other suggestion was just to leave empty brackets (myObject[].myObject2[].c) with out any other markers. But that will help to try request with only one parameter in array.

My opinion, there should be some other way to introduce complex parameters at the ui level, the way different from just simple input fields and submit button.

MaksimOrlov:
And I'm opened for any other suggestions and solution for that issue:)

dilipkrish:
> My opinion, there should be some other way to introduce complex parameters at the ui level, the way different from just simple input fields and submit button.

That way is to use the request body :)

Not sure what the right answer is, how about instead of #ind, we replace it with 0. That way the parameters are atleast accurate for one variation. What do you think?

MaksimOrlov:
I'm agree with you. That can partly solve the issue. 

I will make commit with this changes.

dilipkrish:
Thanks for the PR! 🙇 Sorry it took so long to merge it in

PR #1254, opened 2016-04-09
 PR JSR-303 support for Request Parameters + Fix 1244 Allowable values not displayed in Apidocs + 1231 Externalizing Api Descriptions

jfiala:
Bean Validations API (JSR-303) support for Request Parameters #1227

Issue #1244, opened 2016-03-30
@ApiParam - Allowable values not displayed in Swagger API docs


jfiala:
Version 2.4.0

If e.g. @ApiParam(allowableValues="range[1, 11]") is set, it is never displayed in the API docs (minLength=1, maxLength=11).
It doesn't work directly for a Request parameter:

``` java
@RequestParam @ApiParam(value="description", allowableValues="range[1, 11]") String sample;
``` 

It also doesn't work if the parameter is declared using @ModelAttribute.

For @ApiModelProperty it works as desired:
``` java
@ApiModelProperty(value="description", allowableValues="range[1, 11]")
```
=> will display minLength=1, maxLength=11.

Can you pls check why this is missing for @ApiParam?

jfiala:
In fact, this also affects other ParameterBuilderPlugin / ExpandedParameterBuilderPlugin who also set allowableValues, but it never shows up in the API docs.

jfiala:
This had to be fixed in springfox.documentation.swagger2.mappers.EnumMapper, there was only support for AllowableListValues implemented. Added AllowableRangeValues.

dilipkrish:
- Need to [add range as well here](https://github.com/springfox/springfox/blob/master/springfox-swagger2/src/main/java/springfox/documentation/swagger2/mappers/SerializableParameterFactories.java#L69)

johnsonlegend:
tested version 2.8.0 and 2.9.2, looks like it's not working as expected

hiejulia:
Is it fixed yet ? Same problem happened with `swagger-jersey2-jaxrs`.

dilipkrish:
Yes it should be. Please move to 3.0.0. Also `swagger-jersey2-jaxrs` is not part of this project.

Issue #1228, opened 2016-03-18
Inconsistent attributes for adding description to the API


jfiala:
Version 2.4.0

Current naming scheme for adding a description to the documentation:
@ApiOperation: value
@ApiModel: description // value used for alternative name for the model
@ApiModelProperty: value
@ApiParam: value

IMHO the "description" attribute should be always used for the description of an Operation/Model/ModelProperty.

dilipkrish:
Just FYI, we are using the swagger core annotations. So much of this inconsistency stems from there.

- [@ApiOperation description is from the value](http://docs.swagger.io/swagger-core/current/apidocs/io/swagger/annotations/ApiOperation.html#value())
- [@ApiModelProperty description is from the value](http://docs.swagger.io/swagger-core/current/apidocs/io/swagger/annotations/ApiModelProperty.html#value())
- [@ApiParam description is from the value](http://docs.swagger.io/swagger-core/current/apidocs/io/swagger/annotations/ApiParam.html#value())
- [@ApiModel has a description property](https://github.com/springfox/springfox/blob/master/springfox-swagger-common/src/main/java/springfox/documentation/swagger/schema/ApiModelBuilder.java#L49-L51)

Since we don't control this and its in a dependent library, this will not be fixed

PR #1371, opened 2016-07-02
JSR-303 for Request parameters,  Fix Allowable values not displayed in Apidocs, Externalizing Api Descriptions

jfiala:
@dilipkrish this is the cleaned up version of PR https://github.com/springfox/springfox/pull/1254
- JSR-303 support for Request Parameters #1244 
- Fix Allowable values not displayed in Apidocs #1231
- Externalizing Api Descriptions #1254


springfox-swagger2/src/main/java/springfox/documentation/swagger2/mappers/EnumMapper.java
```diff
@@ -40,12 +41,18 @@
 import static com.google.common.collect.Lists.*;
 
 public class EnumMapper {
+<<<<<<< HEAD
   static ModelImpl maybeAddAllowableValuesToParameter(ModelImpl toReturn, AllowableValues allowableValues) {
+=======
+  static ModelImpl maybeAddEnumValues(ModelImpl toReturn, AllowableValues allowableValues) {
+	  
+>>>>>>> ae0f78d671ca565047555382634db5a98287574c
```
dilipkrish: Seems like this file isn't resolved.

dilipkrish:
I'll polish this change and merge it in later next week! thanks!

jfiala:
@dilipkrish reset EnumMapper with existing springfox-version (it already contains all changes for minimum/maximum of ApiParams)

dilipkrish:
@jfiala This looks good but I'd like to propose some changes and I thought Id run it by you to see what you think.

- Create a new annotation that is specifically for description. `@DescriptionKey` an example would be

```java
    @DescriptionKey("models.SomeClass") // key is springfox.descriptions.models.SomeClass
    class SomeClass {
    }

    // Implicit key is springfox.descriptions.models.SomeOtherClass
    class SomeOtherClass {}
```

Going with the philosophy of this library, make that convention over configuration and have the annotation be optional as in the case of `SomeOtherClass`.

The same analogy can be applied to parameters, properties, operations etc. The other advantage is that we don't need to piggy back on top of the `Api*` annotations that come with problems related to plugin ordering. What do you think? _Caveat:_ I currently don't have the bandwidth to work on it however.

jfiala:
@dilipkrish So in fact this means you only have to annotate each class (model/api) with the @DescriptionKey and we'll do the rest automatically (check if descriptions are present for the class itself, and  for all parameters (models) or methods (apis) inside it.

This should be relatively easy to add, however I'd prefer to still support using the Swagger-description for the keys if someone prefers to use them.
So I'd add a separate Plugin for this feature, so users can use whatever they like or even both...

dilipkrish:
> So in fact this means you only have to annotate each class (model/api) with the @DescriptionKey and we'll do the rest automatically (check if descriptions are present for the class itself, and for all parameters (models) or methods (apis) inside it?

Not quite. We wouldn't need to scan for all the annotations in the class loader at one go. It would be just like how we do things today for description of the parameter we'd only check the parameter, model only the class etc. 

I was suggesting that we unify the documentation of  operations, parameters, models, properties using the natural naming hierarchy (convention) and _overriding_ the keys using the @DescriptionKey on an ass needed basis.

> This should be relatively easy to add, however I'd prefer to still support using the Swagger-description for the keys if someone prefers to use them.

Why? IMO there should be one prescribed way to do something, I also I feel it will confuse the library consumer, especially if the descriptions are not resolved for a given key. Also [as you've pointed out](https://github.com/springfox/springfox/issues/1228), the attributes are not consistently named. 

If we do it this way  we **_won't need to overload_** the `@ApiModel#description` _literal value_ with a the idea of it _being a property key_. 

Lets take an example; 

We have a property file  `swagger-descriptions.properties`

```properties
springfox.descriptions.models.someModel1="Some Model1 via DescriptionKey"
springfox.descriptions.models.someModel2="Some Model2 via DescriptionKey"
springfox.descriptions.models.someModel3="Some Model3 via DescriptionKey"
springfox.descriptions.models.someModel5="Some Model5 via DescriptionKey"
```

... and a collection of classes (just taking the example of documenting models)

```java
   @ApiModel(description="Some Model1 via Api Model Description")
   @DescriptionKey("someModel1")
    class SomeModel1 {
    }

    @DescriptionKey("someModel2") // key is springfox.descriptions.models.SomeModel2
    class SomeModel2 {
    }

    // Implicit key is springfox.descriptions.models.SomeModel3 __NOTE:__ caseing
    class SomeModel3 {}

    @ApiModel(description="Some Model4 via Api Model Description")
    // Implicit key is springfox.descriptions.models.SomeModel4
    class SomeModel4 {}

    @ApiModel(description="Some Model5 via Api Model Description")
    // Implicit key is springfox.descriptions.models.SomeModel5
    class SomeModel5 {}

    // I believe you were suggesting using this as the key
    @ApiModel(description="someModel6") 
    class SomeModel6 { 
    }
```

Will produce the following descriptions

Model | Description
------------ | -------------
SomeModel1 | Some Model1 via DescriptionKey"
SomeModel2 | Some Model2 via DescriptionKey"
SomeModel3 | Some Model3 via DescriptionKey"
SomeModel4 | Some Model4 via Api Model Description
SomeModel5 | Some Model5 via DescriptionKey
SomeModel6 | someModel6

I think it would involve a very minimal change to what you already have working. 

Instead of reading the for e.g. `@ApiModel#description` for key you'd be 
- extracting a key (from DescriptionKey if present, or coming up with a key based on hierarchy of the object)
- Resolving the key if possible

Does that make sense?

jfiala:
Absolutely, I thought it exactly like this (same way for fields / operations)! So in fact users have only to add the @DescriptionKey if they are intending reusing a property key.

We could also allow for using the attribute names only as keys.

An example:
``` java
@ApiModel(description="Some Model1 via Api Model Description")
   @DescriptionKey("someModel1")
    class SomeModel1 {
        private String attribute1;
        private String attribute2;
        @DescriptionKey("reused_key")
        private String attribute3;
    }

    @DescriptionKey("someModel2") // key is springfox.descriptions.models.SomeModel2
    class SomeModel2 {
        private String attribute1;
        private String attribute2;
        @DescriptionKey("reused_key")
        private String attribute3;
    }
```

So one will be able to add the following keys:
| Key | resolves for property | 
| ----- | --------------- |
| SomeModel1.attribute1 | SomeModel1.attribute1 | 
| attribute1 | SomeModel2.attribute1 (Fallback without model prefix) | 
| attribute2 | SomeModel1.attribute2, SomeModel2.attribute2 (only fallback) | 
| reused_key | SomeModel1.attribute3, SomeModel2.attribute3 | 

Using the fallback mechanism for attributes makes it even easier to reuse description across models.
Or does that go too far?

dilipkrish:
Great! I think we're thinking the same thing!

> Using the fallback mechanism for attributes makes it even easier to reuse description across models.
Or does that go too far?

May be a little more effort I would think :) but it would make reusing attribute descriptions easier.

sac10nikam:
Hi,

I have created custom plugins and using it from project. The properties file having key-values is created. but I don't know how to call it from Swagger Controller.

1) 
**ControllerInterface- method**

**@ApiOperation(value = "get_contribution_api_name", notes = "get_contribution_api_notes", responseContainer = "List<Plan>")**
    **@DescriptorKey("get_contribution_api_notes")**
    @ApiResponses(value = { @ApiResponse(code = 200, message = "Success"),
            @ApiResponse(code = 400, message = "Bad Request", response = Response.class),
            @ApiResponse(code = 401, message = "Unauthorized", response = Response.class),
            @ApiResponse(code = 403, message = "Forbidden", response = Response.class),
            @ApiResponse(code = 404, message = "Not Found", response = Response.class),
            @ApiResponse(code = 500, message = "Internal server error occured", response = Response.class) })
    @RequestMapping(value = URIConstants.PLAN_CONTRIBUTION, method = RequestMethod.GET)
    public Response<List<Plan>> **getContribution**(@PathVariable String employeeId,
            @PathVariable String company);

2) property reader class
@Service(value = "environmentConfig")
@Scope(value = "singleton")
@PropertySources
({
@PropertySource(value=**"classpath:/Project/swagger/api-plan-specs/src/resources/swagger-messages.properties**",ignoreResourceNotFound=true)
})
public class **EnvironmentConfig** {

    private static final Logger LOGGER = LoggerFactory.getLogger(EnvironmentConfig.class);
    @Autowired
    private Environment environment;
    private static Environment env;

    @PostConstruct
    public void init() {
        env = environment;
    }

    public static String getProperty(String key) {
        LOGGER.debug("Getting property  :" + key);
        return env.getProperty(key);
    }
}

3) Properties file
swagger-messages.properties

#Swagger errors
get_contribution_api_name=getContribution
get_contribution_api_notes=Gets the list of plan

4) plugins
image

5) Validators
image

Please guide me here. I want to show property values from resource bundle.

dilipkrish:
@jfiala ping!

jfiala:
@dilipkrish can you merge this in and then have a separate issue for the automatic property lookup so I can start from a clean master again, as this is already quite big?

dilipkrish:
Sure. Do you mind if I squash/reword commits as I cleanup and rebase this PR?

jfiala:
@dilipkrish no problem, please do as you like, then we can add the additional functionality (automatic property lookup) in a clean way (new issue/new PR).

Issue #1373, opened 2016-07-05
JSONEditor Error


eruiz:
I have an Spring Boot application and I have followed the [Springfox integration guide](https://springfox.github.io/springfox/docs/snapshot/#springfox-swagger2-with-spring-mvc-and-spring-boot) but the SwaggerUI is not rendering the API.

The console of the browser shows the error below:

> jsoneditor.min.js:7 Uncaught Error: element should be an instance of Element

Note the URL "http://localhost:8080/v2/api-docs" returns the full API spec in JSON format, so sure is a problem in SwaggerUI.

Any idea or help?

Thanks in advance

dilipkrish:
If you've enabled `enableJsonEditor` to true, than turn it back to false.

eruiz:
Yes, but it didn't work. Below the _UiConfiguration_ settings I use:

```
  @Bean
  UiConfiguration uiConfig() {
    return new UiConfiguration(
        "validatorUrl",// url
        "none",       // docExpansion          => none | list
        "alpha",      // apiSorter             => alpha
        "schema",     // defaultModelRendering => schema
        UiConfiguration.Constants.DEFAULT_SUBMIT_METHODS,
        false,        // enableJsonEditor      => true | false
        true);        // showRequestHeaders    => true | false
  }
```

But the JSONEditor .js file is being included in the HTML.

Let me know if you need some info else.

dilipkrish:
The Json editor file will be included, but it shouldn't get activated. try clearing your browser cache

miborra:
Hi! 
I have the same error message when I modify the supportedSubmitMethods parameters. I think that the problem is in file springfox.js because in the line jsonEditor: data.jsonEditor || false, sets jsonEditor to "false" or "true" (string), and jsonEditor isn't  disabled

eight19:
Hi all, I have had the same error. In my case it was because I was passing the element using Jquery selector as below,

`new JSONEditor($('#editor_holder'),{`

where, jquery selector will return as an array of object. In this case I had to send the first value of an array as below fixed the issue and I can see the JSON editor with no errors. 

`new JSONEditor($('#editor_holder')[0],{`

Hope this helps someone. :)

PR #1386, opened 2016-07-13
#1373 - Modified how to set jsonEditor parameter

miborra:
#### What's this PR do/fix?
This PR modifies how to set jsonEditor parameter to solve problem related on #1373 task
No there aren't. To test it you should configure  `UiConfiguration` with `enableJsonEditor` parameter to `false`. In swagger page, the button `Try it!` shouldn't be accesible.
No
#1373


dilipkrish:
👍 thanks for catching that! Do you know if this is a problem in all browsers or just a particular browser?

miborra:
In all browsers

PR #1448, opened 2016-08-12
Update to use swagger2markup version 1.0+

ssevertson:
Swagger2Markup has reached 1.0, with some breaking API changes, and new features, like Confluence Wiki Markup generation. This is a minimal set of changes to use the new version/API.

### DIFF
--- a/springfox-staticdocs/build.gradle
+++ b/springfox-staticdocs/build.gradle
@@ -26,7 +26,8 @@ sourceCompatibility = 1.7
 targetCompatibility = 1.7
 
 dependencies {
-  compile "io.github.robwin:swagger2markup:0.9.2"
+  compile "io.github.swagger2markup:swagger2markup:1.0.1"
+  compile "io.github.swagger2markup:swagger2markup-extension-commons:1.0.0"
   provided "org.springframework:spring-test:${spring}"
   provided libs.clientProvided
   provided libs.springProvided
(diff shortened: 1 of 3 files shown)


ssevertson:
Never mind, didn't realize Swagger2Markup required Java 8, and SpringFox was targeting Java 6. Tests passed locally in my IDE.

dilipkrish:
@ssevertson sadly that is a true statement. I hope to move to jdk 8 but I'm not sure what the right time is. The good news is that using swagger2markup doesn't rely on this library. There are [great docs/tutorials](http://swagger2markup.github.io/swagger2markup/1.0.1/#_spring_boot_and_springfox) by @RobWin that don't require this library of springfox.

Thanks for the PR in any case! 🙇

RobWin:
Yes sry. We decided to switch to Java8 with release 1.0.0

dilipkrish:
I tried upgrading a couple months ago and realized. Thats a good thing tho. No need to stifle innovation based on the lowest common denominator. Thats why Im considering breaking down the this _monolith_ into smaller pieces .. for e.g. swagger-ui, s2m modules into its own repo. Also getting swagger1 module out as well since no-one is using it.

RobWin:
A s2m module isn't even necessary anymore. This would decouple our projects, but they can still be used together.

dilipkrish:
This what I thot but I wasn't sure. I haven't used it in a while.

Issue #1337, opened 2016-06-10
java.sql.Date should translate to format date instead of date-time 


ralfstuckert:
The documentation states that you should use `java.sql.Date` if you want swagger to turn that into type `string` with format `date`. This works fine with springfox-swagger2 version 2.4.0, but with 2.5.0 you get format `date-time`.

This is always reproducable using either using sql.Date directly

`@ApiModelProperty(value = "YYYY-MM-DD, e.g. 2015-01-16")`
`public java.sql.Date mydate;`

or by specifying a dataType:

`@ApiModelProperty(dataType = "java.sql.Date", value = "YYYY-MM-DD, e.g. 2015-01-16")`
`public LocalDate mydate;`

or - as suggested by the documentation - via fixed substitution

`directModelSubstitute(LocalDate.class, java.sql.Date.class)`

dilipkrish:
That's a good catch. Thank u for reporting. 👍🏻

ssevertson:
This is a fairly major issue affecting the usability of dates. It looks like a simple one-line change to springfox.documentation.schema.Types - do you need a PR to get this fixed?

dilipkrish:
@ssevertson sure that would be welcome! :metal:

dilipkrish:
Fixed via #1456

PR #1456, opened 2016-08-17
java.sql.Date should become a "date" in JSON Schema output, according to docs

ssevertson:
Fixes issue #1337.


dilipkrish:
Thanks for the PR! :metal:

PR #1463, opened 2016-08-22
Add 'example' in merged model

wssccc:
Add 'example' in merged model


dilipkrish:
Thanks for catching that!

Issue #1162, opened 2016-02-02
java.sql.Date creates an useless "Date" definition


cbornet:
java.sql.Date fields are correctly mapped to "string" type with "date" format
```json
                "sqldate":
                {
                    "type": "string",
                    "format": "date"
                },
```
but it also creates a useless "Date" object in the definitions:
```json   
     "Date":
        {
            "properties":
            {
                "date":
                {
                    "type": "integer",
                    "format": "int32"
                },
                "day":
                {
                    "type": "integer",
                    "format": "int32"
                },
                "hours":
                {
                    "type": "integer",
                    "format": "int32"
                },
                "minutes":
                {
                    "type": "integer",
                    "format": "int32"
                },
                "month":
                {
                    "type": "integer",
                    "format": "int32"
                },
                "seconds":
                {
                    "type": "integer",
                    "format": "int32"
                },
                "time":
                {
                    "type": "integer",
                    "format": "int64"
                },
                "timezoneOffset":
                {
                    "type": "integer",
                    "format": "int32"
                },
                "year":
                {
                    "type": "integer",
                    "format": "int32"
                }
            }
        }
```
Note : it can be workarounded by setting ```.ignoredParameterTypes(java.sql.Date.class)``` to the docket

dilipkrish:
Yeah its a little annoying. I'll see if I can fix it. May be a pr?

mtalbot:
This was implemented wrong as far as I can tell from the docs. java.sql.Date is supposed to represent a date and not datetime. This was the old behavour. Now both java.util.Date and java.sql.Date report as date times. I've created a pr #1469 to change this to being date instead.

PR #1469, opened 2016-08-26
Fixing java.sql.Date to be a date in swagger instead of date-time

mtalbot:
The original fix in e99f059182ee5ff57b72a670f0761cb3e12b0e68 for #1162 broke the swagger type so that it reported as date-time instead of date.
The unit test from the original fix has been modified to test for date instead of date-time


cbornet:
Duplicate of #1456

dilipkrish:
@mtalbot Thanks for the PR! :metal: I've merged the first PR. Closing this as a duplicate of #1456

Issue #1465, opened 2016-08-23
Type-level @ApiResponses are ignored


sworisbreathing:
Swagger-core allows the `@ApiResponses` annotation to be declared at either the type-level or the method-level, and the [javadoc for `@ApiResponse`](http://docs.swagger.io/swagger-core/current/apidocs/index.html?io/swagger/annotations/ApiResponse.html) says that

> ... class level annotations will be parsed only if an @ApiResponse annotation with the same code is not defined at method level or in thrown Exception

However, it seems that springfox only handles the `@ApiResponses` annotation when it is declared at method-level (even when using `useDefaultResponseMessages(false)`)

I've created a sample app to reproduce the issue. There are two copies of the same API, but one has the `@ApiResponses` annotation on the class declaration and the other has it at only the method-level.

### To reproduce:

---

1. clone https://github.com/sworisbreathing/springfox-issues
2. Start application via mvn spring-boot:run
3. Open URL http://localhost:8080/v2/api-docs (swagger specs) or http://localhost:8080/swagger-ui.html (html version).

Only the successful response is shown for `/type-level-annotation-api` endpoints, while `/method-level-annotation-api` endpoints have both the success and failure responses shown.

---

This issue is somewhat similar to #702, except that I'm using the current release of springfox (2.5.0 as of this writing) and the issue is specific to having the `@ApiResponses` annotation on the class instead of the method. I think there were a few other issues opened requesting a common place to define error responses, and while it's great that there's a place to put this globally (as noted in #1457), this still feels like a bug.

dilipkrish:
You're right there is currently no way to provide controller level globals. Would love for community contributions on this one. 

The work around is to use globals provided via the docket configuration.

marcellodesales:
:( That's sad... Getting into the same exact problem

sworisbreathing:
@dilipkrish: I found a bit of spare time and created a PR for this one. Let me know what you think.

PR #1474, opened 2016-08-30
Merge method-level and class-level @ApiResponses annotations.

sworisbreathing:
#### What's this PR do/fix?
Merges `@ApiResponses` annotations from method-level and type-level (at least, when annotated in the same class). Method-level annotations take priority over type-level annotations.
I have updated `AnnotationsSpec` to verify the behavior when the annotations are defined in the same class. I have not tested subclassing behavior. I have also manually tested against the example project I created for #1465 (https://github.com/sworisbreathing/springfox-issues)
See #1465.
#1465

### DIFF
--- a/springfox-swagger-common/src/main/java/springfox/documentation/swagger/annotations/Annotations.java
+++ b/springfox-swagger-common/src/main/java/springfox/documentation/swagger/annotations/Annotations.java
@@ -36,6 +36,7 @@ import java.util.List;
 import java.util.Set;
 
 import static com.google.common.base.Optional.*;
+import static com.google.common.collect.Lists.*;
 import static org.springframework.core.annotation.AnnotationUtils.*;
 
 public class Annotations {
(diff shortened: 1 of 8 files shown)


springfox-swagger-common/src/main/java/springfox/documentation/swagger/annotations/Annotations.java
```diff
@@ -52,8 +53,16 @@ private Annotations() {
     return fromNullable(findAnnotation(annotated, ApiOperation.class));
   }
 
-  public static Optional<ApiResponses> findApiResponsesAnnotations(Method annotated) {
-    return fromNullable(findAnnotation(annotated, ApiResponses.class));
+  public static List<Optional<ApiResponses>> findApiResponsesAnnotations(AnnotatedElement annotated) {
```
dilipkrish: `List` is monadic in itself, so just a returning a list of ApiResponses would suffice. i.e. don't add it to the list of the `getAnnotation` method returns a null. makes sense? 

It'll make consumption in downstream processes easier, for e.g. we won't need a `if (apiResponses.isPresent()) {...}` block.

dilipkrish:
👍 looks good except for a minor nit! thanks for the PR! Super appreciated!

sworisbreathing:
Okay I've updated the PR to get rid of the `Optional`s in the list. Let me know what you think!

dilipkrish:
Thanks it looks great! Its in!! :bow:

PR #1476, opened 2016-08-31
Fix mistype in documentation

mvpotter:
PR fixes mistype in documentation: bean validation refers JSR-303 not JSR-302


dilipkrish:
Thanks!!

Issue #651, opened 2015-04-06
Feature request: an option for YAML output


who:
It'd be handy to have springfox optionally spit out yaml that could be used in [swagger-editor](http://editor.swagger.io/#/).

dilipkrish:
@who would you be interested in submitting a pr for that?

who:
Yes, I'd be willing to have a go at it.

dilipkrish:
Just to be set the expectations. I think these are the things that need to be done
- [ ] Add yaml to the produces in the controllers ```DefaultSwaggerController``` and ```Swagger2Controller```. Currently it defaults to json
- [ ] Add a configuration option to ```@EnableSwagger(supportsYaml=true)``` and ```@EnableSwagger2(supportsYaml=true``` that we use to register a new message converter in the pipeline. 
- [ ] Make sure the swagger modules are registered with the ```ObjectMapper``` in the new message converter  

This is just an initial outline. Obviously you'll find that reality might be a lot different :) Feel free to add to this discussion, just so that we're aware of the direction you're planning to take. 

/c @adrianbk

adrianbk:
Sounds reasonable.

rLitto:
Hi!
How is the development going? Since editor.swagger.io supports yaml, it would be really interesting to have springfox generate it, since we are considering it to document our microservices

dilipkrish:
@rLitto I think @who is prolly working on it. It's easy enough if you want to take a crack at it.

rLitto:
Would like to. Let me check with my chief if I can make it a spare time
story for our company
Il 19/Mag/2015 19:39, "Dilip Krishnan" <notifications@github.com> ha
scritto:

who:
I've sidelined this because of a project priority change.  Also, swagger-editor can import json files now as well, which somewhat fulfills the initial request in the first place.

ferrarimarco:
I needed this and implemented a Spring HttpMessageConverter for application/x-yaml. I'll see if I can produce a pull request for this feature

marceloverdijk:
I would like yaml output as well. Much more readable than json.

mjwalfredo:
[...]

PR #1488, opened 2016-09-13
Add yml support

apixandru:
This pull request adds basic support for yaml.

Some basic tests added for the mappers, not for the actual endpoint. Also, the content type for the endpoint is still application/json. There is no standard value for content type yaml so far as far as I'm aware of so i left it to application/json for now.

The documentation format is triggered using ?format=
currently only "yml" and "yaml" are acceptable input parameters with everything else defaulting to json. This only applies to the v2 endpoint, not to the v1 which is still always json.

Somewhat related to #651

### DIFF
--- a/springfox-spring-web/src/main/java/springfox/documentation/spring/web/output/MultiFormatSerializer.java
+++ b/springfox-spring-web/src/main/java/springfox/documentation/spring/web/output/MultiFormatSerializer.java
@@ -21,20 +21,22 @@ package springfox.documentation.spring.web.output;
 
 import com.fasterxml.jackson.core.JsonProcessingException;
 import com.fasterxml.jackson.databind.ObjectMapper;
+import org.springframework.http.MediaType;
 import springfox.documentation.spring.web.output.formats.CustomFormatOutputMapper;
 
+import java.util.ArrayList;
 import java.util.Collection;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 
+import static org.springframework.http.MediaType.APPLICATION_JSON;
+
 public class MultiFormatSerializer {
 
-  private static final String FORMAT_JSON = "json";
+  private static final MediaType DEFAULT_MEDIA_TYPE = APPLICATION_JSON;
 
-  private static final String FORMAT_DEFAULT = FORMAT_JSON;
-
-  private final Map<String, ObjectMapper> mappers = new HashMap<String, ObjectMapper>();
+  private final Map<MediaType, ObjectMapper> mappers = new HashMap<MediaType, ObjectMapper>();
 
   public MultiFormatSerializer(List<JacksonModuleRegistrar> modules, Collection<CustomFormatOutputMapper> outputMappers) {
 
@@ -43,30 +45,43 @@ public class MultiFormatSerializer {
       for (JacksonModuleRegistrar each : modules) {
         each.maybeRegisterModule(objectMapper);
       }
-      mappers.put(mapperWrapper.getFormat(), objectMapper);
+      for (MediaType mediaType : mapperWrapper.getFormats()) {
+        mappers.put(mediaType, objectMapper);
+      }
     }
 
   }
 
   public RawOutput toJson(Object toSerialize) {
-    return serialize(toSerialize, FORMAT_JSON);
+    return serialize(toSerialize, DEFAULT_MEDIA_TYPE);
   }
 
-  public RawOutput serialize(Object toSerialize, String format) {
-    String actualFormat = getActualFormat(format);
-    ObjectMapper objectMapper = mappers.get(actualFormat);
+  public RawOutput serialize(Object toSerialize, MediaType mediaType) {
+    ObjectMapper objectMapper = mappers.get(mediaType);
+    if (null == objectMapper) {
+      throw new IllegalArgumentException();
+    }
     try {
       return new RawOutput(objectMapper.writeValueAsString(toSerialize));
     } catch (JsonProcessingException e) {
-      throw new IllegalArgumentException("Could not write " + actualFormat, e);
+      throw new IllegalArgumentException("Failed to serialize object.", e);
     }
   }
 
-  private String getActualFormat(String proposedFormat) {
-    if (mappers.containsKey(proposedFormat)) {
-      return proposedFormat;
+  public MediaType findAvailableMediaType(Collection<MediaType> acceptableMediaTypes) {
+    if (acceptableMediaTypes.isEmpty()) {
+      return DEFAULT_MEDIA_TYPE;
     }
-    return FORMAT_DEFAULT;
+    List<MediaType> sortedMediaTypes = new ArrayList<MediaType>(acceptableMediaTypes);
+    MediaType.sortByQualityValue(sortedMediaTypes);
+    for (MediaType acceptableMediaType : sortedMediaTypes) {
+      for (MediaType availableMediaType : mappers.keySet()) {
+        if (acceptableMediaType.includes(availableMediaType)) {
+          return availableMediaType;
+        }
+      }
+    }
+    return null;
   }
 
 }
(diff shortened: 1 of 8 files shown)


apixandru:
The swagger editor seems to be happy with the output

image

apixandru:
Maybe it's not a good idea to add the dependency like I did right now directly to the web project because we will always be importing it.

If we add it to dependencies in the client provided section, we will get java.lang.ClassNotFoundException: com.fasterxml.jackson.dataformat.yaml.YAMLFactory 

Maybe another solution would be a spring boot autoconfiguration class to add the yaml mapper only if it finds the class at runtime.

springfox-swagger2/src/main/java/springfox/documentation/swagger2/web/Swagger2Controller.java
```diff
@@ -63,29 +63,30 @@
...
+  private MultiFormatSerializer multiFormatSerializer;
 
   @ApiIgnore
   @RequestMapping(value = "${springfox.documentation.swagger.v2.path:" + DEFAULT_URL + "}",
       method = RequestMethod.GET, produces = { APPLICATION_JSON_VALUE, HAL_MEDIA_TYPE })
   public
   @ResponseBody
-  ResponseEntity<Json> getDocumentation(
+  ResponseEntity<RawOutput> getDocumentation(
+      @RequestParam(value = "format", required = false) String format,
```
dilipkrish: Why not make this driven by the accept header?

`@RequestHeader(value = "Accept",  defaultValue="application/json") String format,` and add the supported media types to the operations `@RequestMapping`.

gradle/dependencies.gradle
```diff
@@ -51,6 +51,9 @@ ext {
                   "com.fasterxml.jackson.core:jackson-databind:${jackson}",
                   "joda-time:joda-time:$joda"
           ],
+          client        : [
```
dilipkrish: May be make this part of clientProvided dependencies.

springfox-spring-web/src/main/java/springfox/documentation/spring/web/output/MultiFormatSerializer.java
```diff
@@ -0,0 +1,77 @@
...
+    ObjectMapper yamlMapper = new ObjectMapper(new YAMLFactory());
+    ObjectMapper jsonMapper = new ObjectMapper();
+
+    for (JacksonModuleRegistrar each : modules) {
+      each.maybeRegisterModule(yamlMapper);
+      each.maybeRegisterModule(jsonMapper);
+    }
+
+    mappers.put(FORMAT_YML, yamlMapper);
+    mappers.put(FORMAT_YAML, yamlMapper);
```
dilipkrish: Conditionally add the formats if `YAMLFactory` is available in the class path

springfox-spring-web/src/main/java/springfox/documentation/spring/web/output/MultiFormatSerializer.java
```diff
@@ -0,0 +1,77 @@
...
+    return serialize(toSerialize, FORMAT_JSON);
+  }
+
+  public RawOutput serialize(Object toSerialize, String format) {
+    String actualFormat = getActualFormat(format);
+    ObjectMapper objectMapper = mappers.get(actualFormat);
+    try {
+      return new RawOutput(objectMapper.writeValueAsString(toSerialize));
+    } catch (JsonProcessingException e) {
+      throw new IllegalArgumentException("Could not write " + actualFormat, e);
```
dilipkrish: throw a [405](https://http.cat/415) exception here so that its clear to clients and does not come through as a 500 error

dilipkrish:
@apixandru First off thanks for the awesome PR! :metal: 

Added a couple of suggestions ☝️ o/w looks great!

springfox-swagger2/src/main/java/springfox/documentation/swagger2/web/Swagger2Controller.java
apixandru: The main reason is that when I usually "demo" the swagger docs i do it from the browser, not via a web client where I can set additional headers. 

The second reason why I did this was because yaml doesn't have any internet media type registered yet. This second reason is why i didn't change the produces header here as well, there is no standard for yaml as far as I'm aware of.

dilipkrish: Thats why I suggested a default value. Since many browsers will set the header to `*/*` or equivalent, we can override it to `application/json`.

About the format parameter, Im hesitant to tunnel formats via query string, I'd rather go with a slew of accepted (not a standard) media types for yaml `application/yaml`, `application/yml`, `text/yaml`  etc. [as suggested here](http://stackoverflow.com/questions/332129/yaml-mime-type).

apixandru:
I added the conditional dependency but I did it using spring boot's autoconfiguration. Not sure if that's what you had in mind. Do you have another suggestion on the conditional dependency?

gradle/dependencies.gradle
```diff
@@ -44,15 +45,14 @@ ext {
...
           ],
           clientProvided: [
                   "javax.servlet:javax.servlet-api:$servlet",
                   "com.fasterxml.jackson.core:jackson-core:${jackson}",
                   "com.fasterxml.jackson.core:jackson-databind:${jackson}",
-                  "joda-time:joda-time:$joda"
-          ],
-          client        : [
                   "com.fasterxml.jackson.dataformat:jackson-dataformat-yaml:${jackson}",
+                  "joda-time:joda-time:$joda",
```
apixandru: I kept it as an optional dependency, I removed the client part and added the yaml as another optional dependency. it shows this line as green because I added the yaml dependency before it.

apixandru:
I kind of figured that the spring boot autoconfiguration was not the intended solution, but I couldn't think of another way to configure optional custom providers. If you have a better idea I'm happy to have a go at it.

The latest swagger editor doesn't have the option to generate a spring mvc application anymore so it's not really easy for me to test what happens when i call it without spring boot. I'll try to download an older version to test it out at some point when I will have some more free time.

I added the code for the format as Accept header although I think it would have been better as a parameter due to the fact that you would have more extensibility. Maybe something like /api-docs.yml or api-docs.json as endpoints would have been a better option in my opinion.

dilipkrish:
It seems to be still there swagger_editor

apixandru:
What I meant is that there used to be Spring Boot and Spring MVC there, now they only have Spring (which is spring boot). I know it works with spring boot but i'm not sure what happens without boot. I suspect that it should be fine because the YML autoconfiguration class won't be loaded but I'm not 100% sure.

dilipkrish:
Have you tried forking [the springfox-demos repo](https://github.com/springfox/springfox-demos)?

apixandru:
I didn't know about that. Thanks.
I also took a look at the swagger editor source code and found out that they were using the swagger-codegen for the editor and the spring-mvc is also still available as an option in it. (but not on the ui)

Anyway, I used the springfox-demos (both the xml version and the java version) and they work as expected in that the spring boot autoconfiguration is not picked up so there is no missing class exception. They work for json or no value but not for yaml

This means that the pull request does enable yaml but just for spring-boot, there is no regression on the regular spring mvc configuration.

One more point that I'd like to bring up, because of the accept header, there would be some extra configuration on the part of the user to enable the nonstandard "application/yaml" to go through the spring converter. Currently, you get a 406 Not Acceptable. This means that if you debug it, the code executes successfully, it's just that spring doesn't know how to use that return value.

phedoreanu:
Great work @apixandru 👍 

@dilipkrish are you going to merge [it](1488)? It would be great to have yml support.

apixandru:
I have deleted this repository a while back so I can't submit any more commits but I have opened https://github.com/springfox/springfox/pull/2102 instead

Issue #1497, opened 2016-09-27
springfox-staticdocs : Update to last swagger2markup version


orevial:
Currently `springfox-staticdocs` module depends on an old version of `swagger2markup` (`io.github.robwin:swagger2markup:0.9.2`), but it has been released in new versions, the last one being **1.0.1**. A lot of code has changed in `swagger2markup` between the 2 versions but using last lib version would allow for more customization of the `Swagger2MarkupResultHandler`configuration.

PR #1498, opened 2016-09-27
Add pathsGroupedBy configuration

orevial:
This PR adds a missing "withPathsGroupedBy" configuration method to enable configuration of Swagger paths (e.g. controllers methods can get sorted by "TAGS" instead of default "AS_IS" config).
This PR will be probably be useless once swagger2markup gets updated to its last version (see issue #1497) but in the meantime this should be a simple and convenient add.


dilipkrish:
🤘 thanks!

PR #1517, opened 2016-10-12
Added bean validation support for @DecimalMin/@DecimalMax and made the min/max values explicit

jamesbassett:
Added bean validation support for @DecimalMin/@DecimalMax. Code is heavily based on existing `MinMaxAnnotationPlugin`.

I was tempted to change the default min/max values from `Double.MAX_VALUE` to reflect the actual property type (e.g. an Integer should use `Integer.MAX_VALUE` for example), but that is probably a separate issue.

### DIFF
--- a/springfox-bean-validators/src/main/java/springfox/bean/validators/plugins/DecimalMinMaxAnnotationPlugin.java
+++ b/springfox-bean-validators/src/main/java/springfox/bean/validators/plugins/DecimalMinMaxAnnotationPlugin.java
@@ -83,14 +83,12 @@ public class DecimalMinMaxAnnotationPlugin implements ModelPropertyBuilderPlugin
     } else if (decimalMin.isPresent()) {
       LOG.debug("@DecimalMin detected: adding AllowableRangeValues to field ");
       DecimalMin min = decimalMin.get();
-      // use Max value until "infinity" works
-      myvalues = new AllowableRangeValues(min.value(), !min.inclusive(), Double.toString(Double.MAX_VALUE), false);
+      myvalues = new AllowableRangeValues(min.value(), !min.inclusive(), null, null);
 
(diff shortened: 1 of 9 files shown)


jamesbassett:
@dilipkrish So the Circle CI container ran out of memory and gradle died - is this a common build issue with the project? (I don't think my change would have caused it!)

dilipkrish:
Not directly. Since upgrading to gradle 3 the daemon is automatically active. So I disabled it in master. May be rebasing against master will fix your build?

jamesbassett:
I just squashed my commits to trigger the build again....and it works! haha

jamesbassett:
I'm going to update this PR to support exclusiveMin/exclusiveMax as well, as that info is available for `@DecimalMin/@DecimalMax`.

dilipkrish:
Looking forward to it

jamesbassett:
Ok, should be good to go. I'm working on a PR for swagger2markup now to incorporate this into the generated asciidoc as well :)

springfox-bean-validators/src/main/java/springfox/bean/validators/plugins/DecimalMinMaxAnnotationPlugin.java
```diff
@@ -0,0 +1,99 @@
+/*
+ *
+ *  Copyright 2015 the original author or authors.
```
dilipkrish: Would you mind changing the copyright dates to 2016?

jamesbassett:
I hope you don't mind, but I bundled my changes for #1521 (removing defaults for min/max) into this PR as it touched the same code.

Did you need me to address the codacy issues (they're all in the tests, and will also be issues in the existing codebase as I just copied the existing min/max tests)?

dilipkrish:
Thanks for the PR! 🙇

Issue #1525, opened 2016-10-17
2.6.0 breaks the swagger documentation link


apixandru:
In spring 2.6.0 springfox.documentation.swagger.v2.path is ignored.

This has already been fixed by https://github.com/springfox/springfox/commit/383162c8e6281cbaa6da0a0c7c787ec05d380b30

I just created this issue for tracking purposes only, however it would probably a good idea to have unit tests covering this to ensure it doesn't happen in the future.

PR #1526, opened 2016-10-17
Add test for issue #1525

apixandru:
Add test for issue #1525


apixandru:
this doesn't actually test that the file was actually loaded...

PR #1576, opened 2016-11-22
Added support for composed bean validation constraints

jamesbassett:
#### What's this PR do/fix?
Enables support for [composed bean validation constraints](https://docs.jboss.org/hibernate/validator/5.0/reference/en-US/html/validator-customconstraints.html#validator-customconstraints-compound). 

For example, you can define a `@Percent` constraint that is annotated with `@DecimalMin("0.0")` and `@DecimalMax("100.0")` and reuse it everywhere you have a percent field.

The current plugins only support the annotations being present on the field/getter itself, not via another constraint (meaning you'd have to write a plugin for each composed constraint, which is not ideal).
Please refer to `BeanValidatorsSpec`.
This solution makes use of Spring's `AnnotationUtils.findAnnotation()`, which takes care of finding annotations (even when they're applied to other annotations), as well as preventing infinite recursion. Being a Spring extension, I thought this was more ideal than writing my own solution!

* Using Spring's AnnotationUtils to find annotations
* Added new `extractAnnotation()` method to BeanValidators as it's common to all plugins (left existing methods for backwards compatability)
* Refactored all plugins to use `extractAnnotation()` - so they now all support composed constraints

### DIFF
--- a/springfox-bean-validators/src/main/java/springfox/bean/validators/plugins/BeanValidators.java
+++ b/springfox-bean-validators/src/main/java/springfox/bean/validators/plugins/BeanValidators.java
@@ -18,16 +18,16 @@
  */
 package springfox.bean.validators.plugins;
 
-import com.fasterxml.jackson.databind.introspect.AnnotatedMember;
 import com.fasterxml.jackson.databind.introspect.BeanPropertyDefinition;
-import com.google.common.annotations.VisibleForTesting;
 import com.google.common.base.Optional;
-import com.google.common.collect.FluentIterable;
 import org.springframework.core.Ordered;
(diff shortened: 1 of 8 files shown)


dilipkrish:
Thanks for the pr @jamesbassett will polish and pull it in in a bit.

jamesbassett:
Hi @dilipkrish, just wondering when we can get this merged? We'd love to use this :)

dilipkrish:
Sorry for the delay, I've been on vacation and haven't had time work on it.

jamesbassett:
No worries, thanks for the update. Hope you had a good break :)

dilipkrish:
Thank you again! 🙇

PR #1589, opened 2016-11-30
fix ApiParam hidden attribute

defshine:
Attribute of @ApiParam  don't work in this situation:

public Page<User> listUser(@ModelAttribute UserQueryRequest request,
                                    @ApiParam(hidden = true)
                                    @PageableDefault(sort = {"createdTime"}, direction = Sort.Direction.DESC)
                                            Pageable pageable) { }


dilipkrish:
@defshine Thanks so much for the PR! Much appreciated! 🙇

Issue #1592, opened 2016-12-01
OAuth2 clientSecret shouldn't be required for implicit flow


gonzalad:
I'm using OAuth2 implicit flow and there's an issue when I set clientSecret to null :  
the URL generated from swagger js for starting the OAuth2 authentication flow contains an undefined client_id (on serverside, clientId is not empty).

Sample URL :  
http://localhost:9080/oidc/idp/authorize?response_type=token&redirect_uri=http%3A%2F%2Flocalhost%3A7777%2Fwebjars%2Fspringfox-swagger-ui%2Fo2c.html&realm=undefined&client_id=undefined&scope=openid&state=oauth2

The issue appears to be in https://github.com/springfox/springfox/blob/2.6.1/springfox-swagger-ui/src/web/js/springfox.js#L71 : OAuth2 is initialized only if clientSecret is set.

What version of the library are you using? 2.6.0 (but the LOC in question is also in 2.6.1)

Thanks,
Adrian

# Additional resources
Here's the JSON returned from /swagger-resources/configuration/security :
```
{  
   "clientId":"FFWtj97yGiRqdQ",
   "realm":"oauth2",
   "appName":"SCIM Service",
   "apiKeyVehicle":"header",
   "scopeSeparator":" "
}
```

Here's a fragment from my Spring Boot Config (the important thing being the SecurityConfiguration : clientSecret is set to null in the constructor): 
```
@Configuration
public class SwaggerConfig {

    public static final String SECURITY_SCHEME_OAUTH2 = "oauth2";

    private static final String OAUTH2_SCOPE_SEPARATOR = " ";

    @Autowired
    private Environment environment;

    @Bean
    public Docket myApi(ServletContext servletContext, ResourceServerConfigurer resourceServerConfigurer,
            OAuth2AutoConfiguration oAuth2AutoConfiguration, List<SecurityScheme> securitySchemes,
            List<SecurityContext> securityContexts) {
        return new Docket(DocumentationType.SWAGGER_2)
                .host(host())
                .securitySchemes(securitySchemes != null ? securitySchemes : new ArrayList<SecurityScheme>())
                .securityContexts(securityContexts)
                .pathProvider(pathProvider(servletContext))
                .groupName("org.mycompany.iam.im.my")
                .apiInfo(myApiInfo())
                .select()
                .paths(regex("/(Groups|Users|Me|Schemas|ResourceTypes|ServiceProviderConfig).*"))
                .build();
    }

    @Bean
[...]

PR #1593, opened 2016-12-01
OAuth2 not initialized when clientSecret undefined

gonzalad:
#### What's this PR do/fix?

fix for #1592 
OAuth2 clientSecret shouldn't be required for implicit flow

Manually tested, the issue is in js code

See #1592 

When clientSecret is set to null on the server-side,
the client-side doesn't initialize any OAuth2 values.
Problem is that clientSecret is not used in some oauth2
flow (i.e. implicit flow).

The URL generated to initiate OAuth2 authentication flow
contains an undefined client_id parameter.

### DIFF
--- a/springfox-swagger-ui/src/web/js/springfox.js
+++ b/springfox-swagger-ui/src/web/js/springfox.js
@@ -68,7 +68,6 @@ $(function() {
 
     function oAuthIsDefined(security) {
       return security.clientId
-          && security.clientSecret
           && security.appName
           && security.realm;
     }


dilipkrish:
Do you think the check should discriminate between flows when checking for oAuth required fields? This change would make it not required for other flows as well. ___DISCLAIMER:___ I'm not familiar with all the flows.

gonzalad:
Hi @dilipkrish !

I'd say to keep it simple : don't do checks. 
Your user will have a clear error returned by the Authorization Server in any way.

I don't use all OAuth flows, so I can't tell you which flow requires a client_secret and which not.

From my knowledge : 
 * authorization code flows requires a client_secret (but this flow should never be used by a front end app - so I cannot understand why it's supported by swagger-ui as is).
 * implicit flow (which don't requires it).

I'd even say that if you're initiating an authentication flow from a front-end app (our case with swagger), then the front-end app is a *Public Client* and should theorically never have a client_secret.
See chapter '2.1.  Client Types' from https://tools.ietf.org/html/rfc6749#page-14.

mmsinclair:
@dilipkrish any idea when you might accept this PR?

Btw, I agree with @gonzalad that you shouldn't be enforcing that all those parameters are set - we have a public client and our identity server only expects _client_id_ to be set. Luckily it ignores the other parameters so I can set them to any string value.

dmzaytsev:
@dilipkrish yes,  public clients (which usually use implicit flow) has no *client_secret*
I agree with @mmsinclair and @gonzalad *client_secret* shouldn't be forced

dilipkrish:
Thanks for weighing in @dmzaytsev, @mmsinclair and @gonzalad. 👍

Issue #1641, opened 2017-01-20
"integer" isn't seen as a valid type by ModelRef


ahatzz11:
version `2.6.1`

I'm trying to create a `ModelRef` to use in a `Header` so I can have headers in my responses. 

```
private ModelRef integerModel = new ModelRef("integer");
private Header xRateLimitLimit = new Header("X-RateLimit-Limit", "The defined maximum number of requests available to the consumer for this API.", integerModel);

private ArrayList<ResponseMessage> responseList = newArrayList(
        new ResponseMessageBuilder()
            .code(400)
            .headersWithDescription(defaultHeaders)
            .message("Bad Request").responseModel(responseCodeModel).build());
```

I am making a separate `ModelRef` and `Header` because I plan on using them quite a few times, and it's easier than creating a new one over and over. 

When I do this, I expect `"type":"integer","format":"int32"` in my swagger doc, instead I am getting `"$ref":"#/definitions/integer"`. 

I have used `"number"`, `"long"`, and `"string"`, and everything works as expected. Instead of using `$ref`, it uses `type`, as expected. This doesn't happen with `"integer"`. This should work according to the [OpenAPI-Specification](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/1.2.md#431-primitives). The common name `"integer"` should have a type of `integer` and a format of `int32`.

--- 
example:

```
    private ModelRef integerModel1 = new ModelRef("integer");
    private ModelRef integerModel2 = new ModelRef("long");
    private ModelRef integerModel3 = new ModelRef("string");
``` 

A different model for each header.

```
    private Header xRateLimitLimit = new Header("X-RateLimit-Limit", "The defined maximum number of requests available to the consumer for this API.", integerModel1);
    private Header xRateLimitRemaining = new Header("X-RateLimit-Remaining", "The number of calls remaining before the limit is enforced and requests are bounced.", integerModel2);
    private Header xRateLimitReset = new Header("X-RateLimit-Reset", "The time, in seconds, until the limit expires and another request will be allowed in. This header will only be present if the limit is being enforced.", integerModel3);
```

I then put those headers into a map...

```
    private Map<String, Header> defaultHeaders = new HashMap<String, Header>(){{
        put("xRateLimitLimit", xRateLimitLimit);
        put("xRateLimitRemaining", xRateLimitRemaining);
        put("xRateLimitReset", xRateLimitReset);
    }};
```

...so that I can use it in `ResponseMessageBuilder`...

```
[...]

PR #1642, opened 2017-01-20
Change "int" type to "integer" to match Open API Specification

ahatzz11:
This PR fixes the integer type to match the Open API Specification. It should be `"integer"`, the current project uses `"int"`. 

This PR also includes fixes to relevant unit tests. There are some failing tests in `springfox-swagger1` and `springfox-swagger2`, but those seemed to failing before I changed anything.

Relevant Issue: https://github.com/springfox/springfox/issues/1641

[The primitives list](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/1.2.md#431-primitives) for Open API shows that the common name should be `"integer"`.


dilipkrish:
Closing this as its [working as designed](https://github.com/springfox/springfox/issues/1641#issuecomment-274194060)

Issue #1507, opened 2016-10-04
Broken basePath with AbstractPathProvider in version 2.5.0 of springfox-swagger2


danielbcorreia:
- Version 2.5.0 (same in 2.6.0)
- Bug: AbstractPathProvider extension point is broken for swagger2 due to this code block in Swagger2Controller:

``` java
    Swagger swagger = mapper.mapDocumentation(documentation);
    if (isNullOrEmpty(swagger.getHost())) {
      final UriComponents uriComponents = componentsFrom(servletRequest);
      swagger.basePath(Strings.isNullOrEmpty(uriComponents.getPath()) ? "/" : uriComponents.getPath());
      swagger.host(hostName(uriComponents));
    }
```

Here you can see that even if the new path is getting to the Swagger object (line 1), it gets overwritten when "getHost" is checked. According to the swagger spec the host isn't even required, so this kind of assumptions should not be made.

On the other hand, the basePath is overwritten when a host is not provided which is the main reason for this bug.

The last version that we know that works is 2.3.1.

Sample implementation of AbstractPathProvider:

``` java
    class BasePathAwareRelativePathProvider extends AbstractPathProvider {
        private String basePath;

        public BasePathAwareRelativePathProvider(String basePath) {
            this.basePath = basePath;
        }

        @Override
        protected String applicationPath() {
            return basePath;
        }
        ...
    }
```

danielbcorreia:
This is the commit that breaks this:

https://github.com/springfox/springfox/commit/a1ceac24e7cd0c646a6217391baa9d8a44fdb324

dilipkrish:
@danielbcorreia this was introduced to allow the application to work correctly in a load balanced environment. Not sure what is breaking here, the fact that the host name isn't what was supplied or that the host name is incorrect?

danielbcorreia:
Specifying the hostname is optional. It's fine to assume the hostname from the current servlet context as it was done before, even though that field is optional in the swagger spec, but overwriting the basePath when the hostname is not supplied creates a lot of issues.

dilipkrish:
Issues like? Could you provide a breaking test or an example?

danielbcorreia:
Sample entrypoint (springboot):

```java

@EnableSwagger2
@SpringBootApplication
public class SampleAPI {

    public static void main(String[] args) {
        SpringApplication.run(SampleAPI.class, args);
    }

    @Value("${api.basePath:}")
    public String apiBasePath;

    @Bean
[...]

PR #1699, opened 2017-03-01
Make Swagger2Controller respect basePath even if a host is not set

### LINKED ISSUE #1435: Setting a Custom basePath Requires Setting a Static Host in 2.5
I have a Spring MVC application that can be reached URLs like:

```
http://my.host.com:8080/my-context/controller-name
```

In order to simplify access to this application, I'm using Apache as a reverse proxy.  As a result of this, my application can be reached (through apache) by URLs like:

```
http://my.host.com/controller-name
```

This creates problems for Springfox because by default it it includes the servlet context in API URLs.  It assumes that the way to reach my controller endpoint is:

```
http://my.host.com/my-context/controller-name
```

This 404s.  So, in order to fix that I created a custom PathProvider that removes the servlet context from the URL.  This works fine in Springfox 2.4 but appears to have been broken in 2.5 by #1217.  Now the custom path generated by the PathProvider is ignored unless I explicitly set a host as well.  I can't set a host because my application is served under multiple domains so I want Springfox to continue to dynamically determine the host.

asdcdow:
This fixes #1435 and #1507

### DIFF
--- a/springfox-swagger2/src/main/java/springfox/documentation/swagger2/web/HostNameProvider.java
+++ b/springfox-swagger2/src/main/java/springfox/documentation/swagger2/web/HostNameProvider.java
@@ -18,7 +18,6 @@
  */
 package springfox.documentation.swagger2.web;
 
-import org.springframework.util.StringUtils;
 import org.springframework.web.servlet.support.ServletUriComponentsBuilder;
 import org.springframework.web.util.UriComponents;
 import org.springframework.web.util.UrlPathHelper;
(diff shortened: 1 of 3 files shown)


dilipkrish:
@asdcdow the PR has some issues ☝️

springfox-swagger2/src/main/java/springfox/documentation/swagger2/web/HostNameProvider.java
```diff
@@ -31,8 +32,8 @@ public HostNameProvider() {
     throw new UnsupportedOperationException();
   }
 
-  static UriComponents componentsFrom(HttpServletRequest request) {
-    ServletUriComponentsBuilder builder = ServletUriComponentsBuilder.fromServletMapping(request);
+  static UriComponents componentsFrom(final HttpServletRequest request, final String basePath) {
```
asdcdow: I didn't fix this complexity issue since I don't think I changed the complexity of this method.  Let me know if you want me to fix this.

Although, I'm a little confused as to how it ends up a complexity of 800.  My understanding is that the complexity number is calculated based  on the number of paths through a method and I cannot figure out how there are 800 paths through that method.

dilipkrish:
@asdcdow thanks for this PR! 🙇

PR #1702, opened 2017-03-02
Make ApiResourceController methods public

### LINKED ISSUE #1701: Make ApiResourceController methods public
[ApiResourceController](https://github.com/springfox/springfox/blob/master/springfox-swagger-common/src/main/java/springfox/documentation/swagger/web/ApiResourceController.java) is a `public` class with no `public` methods in it. It would be useful to make the 3 actions that it defines `public`.

In my use case, I need to put certain additional annotations on these methods to make them publicly visible to our routing library. I'd like to either be able to inherit from the class or delegate to the controller instance from my own controller to achieve this, but with the code as is, I have to define my class within the `springfox.documentation.swagger.web` package.

psyho:
#### What's this PR do/fix?

Closes #1701 

This PR only changes the visibility of 3 methods, it does not affect the functionality of the project in any other way.

Context described in the issue. For convenience, I'm quoting it here:

> ApiResourceController is a public class with no public methods in it. It would be useful to make the 3 actions that it defines public.
> 
> In my use case, I need to put certain additional annotations on these methods to make them publicly visible to our routing library. I'd like to either be able to inherit from the class or delegate to the controller instance from my own controller to achieve this, but with the code as is, I have to define my class within the springfox.documentation.swagger.web package.

#1701

### DIFF
--- a/springfox-swagger-common/src/main/java/springfox/documentation/swagger/web/ApiResourceController.java
+++ b/springfox-swagger-common/src/main/java/springfox/documentation/swagger/web/ApiResourceController.java
@@ -49,21 +49,21 @@ public class ApiResourceController {
 
   @RequestMapping(value = "/configuration/security")
   @ResponseBody
-  ResponseEntity<SecurityConfiguration> securityConfiguration() {
+  public ResponseEntity<SecurityConfiguration> securityConfiguration() {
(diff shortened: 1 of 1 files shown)


dilipkrish:
@psyho thank you!

PR #1717, opened 2017-03-09
Adding vendorExtensions in apiKey class

cfernandezh:
(1636)


dilipkrish:
@cfernandezh thank you so much. 🙇 

I did modify your work a little to use the `VendorExtensions` that were already available, instead of using a `Map<String, Object>` to hold an object. 

Apologies for not mentioning it in the original feature. This is pretty valuable so thank you!

PR #1758, opened 2017-04-06
Use double backtick to prevent miss parsing

naxhh:
#### What's this PR do/fix?
Backticks don't work ok when not followed by a space. Double backticks do.

No unit test, you may want to ensure that ASCIIDocs parses double backtick correctly maybe? I don't know how ASCIIdocs works


dilipkrish:
Thanks for fixing the docs @naxhh

PR #1776, opened 2017-04-18
Fix Swagger version error in documentation

ersinciftci:
In `Customizing the swagger endpoints` section, second line in the first table has Swagger version `2.0`, but it should be `1.2`.


dilipkrish:
Thanks @ersinciftci!

PR #1806, opened 2017-05-09
Update PropertySourcedRequestMappingHandlerMapping.java

OzgaRobert:
mappingPath method used a unresolved value as property lockup key.
So by default we were looking for "${springfox.documentation.swagger.v2.path}" property instead of "springfox.documentation.swagger.v2.path"

Because of this bug it was impossible to use springfox.documentation.swagger.v2.path property to change default mapping.


dilipkrish:
Thanks for the finding that and for the PR! 🙇

Issue #521, opened 2014-12-05
Support for rendering exception response statuses in response codes


dilipkrish:
``` java
    @RequestMapping
    public void someMethod() throws ResourceNotFoundException {...}

   @ExceptionHandler(ResourceNotFoundException)
   @ResponseStatus(HttpStatus.NOT_FOUND)
   void throw404() {...}

```

should generate 404 as a response status for declared exceptions

omayevskiy:
:+1:

GitVhaos:
This feature would be awesome. Should be also supported when declaring @ExceptionHandler in "@ControllerAdvice" annotated classes.

apomelov:
Awesone! Any plans for this issue?

evgeny-konovalov:
+1

alex-under:
+1

ret0:
+1

ianaz:
+1

rmasgutov:
+1

zacyang:
Hi all,
I am currently working on a project which requires this feature to generate swagger spec based on exception handler.
Therefore, I would like to try to fix this if no one picked up this issue yet. 
Also, if there's any previous discussions regarding  this issue regarding the potential approach, it would be helpful.
Thanks.

dilipkrish:
Not sure that it is easy. So any approach you take, if you solve the problem we'd be glad to accept your contribution. Perhaps easier to do via some (custom or otherwise) annotations. At a high level the problem with controller exception handler is that there is no way to know if your controller does not use checked exception without additional annotations.

benweet:
I couldn't find a way to the `org.springframework.web.method.HandlerMethod` (the `java.lang.reflect.Method` either) from the RequestMappingContext/OperationContext. Is it possible to expose one or the other so that we can work with it in our plugins ?

dilipkrish:
@benweet that is intentionally protected from being exposed. Would you mind explaining how you're using the method and we can see if we can abstract that on one of those contexts

benweet:
@dilipkrish I would have a class that implements the `OperationBuilderPlugin` interface and, inside the `apply(OperationContext operationContext)` method, I would list the operation exception via `java.lang.reflect.Method.getExceptionTypes()` (which I can't access at the moment). For each exception type, I would add a ResponseMessage to the OperationContext.

dilipkrish:
@benweet isnt that implementation details that is better supported by adding support for `@ExceptionHandler`?

[...]

PR #1817, opened 2017-05-15
Support javadoc based document generation => Issue 1691

rgoers:
This is a fix for issue 1691. It allows method comments and parameter descriptions to be pulled from Javadoc. It also will copy the required attribute from the Spring annotation to the ApiParam annotation if it is ommitted on the ApiParam.  This feature works by running the Javadoc doclet during the build. The doclet will generate a properties file that is then read during execution. The doclet can be configured in a Maven build as

      <plugin>
        <groupId>org.apache.maven.plugins</groupId>
        <artifactId>maven-javadoc-plugin</artifactId>
        <version>2.10.4</version>
        <executions>
          <execution>
            <goals>
              <goal>javadoc</goal>
            </goals>
            <phase>process-classes</phase>
            <configuration>
              <doclet>springfox.documentation.spring.web.DocletGenerator</doclet>
              <docletArtifact>
                <groupId>io.springfox</groupId>
                <artifactId>springfox-spring-web</artifactId>
                <version>2.7.0-SNAPSHOT</version>
              </docletArtifact>
              <additionalparam>
                -classdir ${project.build.outputDirectory}
              </additionalparam>
              <sourcepath>${project.build.sourceDirectory}</sourcepath>
              <subpackages>com.example</subpackages>
              <useStandardDocletOptions>false</useStandardDocletOptions>
            </configuration>
          </execution>
        </executions>
      </plugin>


rgoers:
FYI - I am not familiar with Gradle at all or the testing mechanisms being used in Springfox so it isn't clear to me how to create unit tests for this. I have tested this in my Maven-based projects.

dilipkrish:
Hey this is a great approach! thanks for working on this. However, I'd like to suggest a few things to pull this in. Let me look at it in detail and provide feedback!

rgoers:
Have you had a chance to review this yet? I have several teams with projects that would like to take advantage of this.

springfox-spring-web/src/main/java/springfox/documentation/spring/web/plugins/JavaDocPropertiesReader.java
```diff
@@ -0,0 +1,77 @@
...
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.springframework.stereotype.Component;
+
+/**
+ *
+ */
+@Component
+public class JavaDocPropertiesReader {
```
dilipkrish: We don't need to manually read property files here. Spring has a much easier way to read properties loaded into the environment. So for e.g. In your configuration of this module you might do something like this

```
@Configuration
@PropertySource("classpath:springfox.javadoc.properties")
public class JavadocModuleConfiguration {
}
```

and you could easily access these properties using the spring `Environment`. Infact we already do something very similar in [`ApiModelPropertiesBuilder`](https://github.com/springfox/springfox/blob/master/springfox-swagger-common/src/main/java/springfox/documentation/swagger/schema/ApiModelPropertyPropertyBuilder.java#L63) using [`DescriptionResolver`](https://github.com/springfox/springfox/blob/master/springfox-spring-web/src/main/java/springfox/documentation/spring/web/DescriptionResolver.java). See here for [an example](https://github.com/springfox/springfox/blob/master/springfox-swagger-common/src/main/java/springfox/documentation/swagger/schema/ApiModelProperties.java#L103-L118).

Secondly the property file need not be in the META-INF directory. It can be in the resources directory as well, which makes it easy for spring to find

springfox-spring-web/build.gradle
```diff
@@ -11,6 +11,7 @@ ext {
 
 dependencies {
   compile "org.reflections:reflections:$reflections"
+  compile files("${System.properties['java.home']}/../lib/tools.jar")
```
rgoers: If you want this in a different repo that's fine. I can create a new repo in my github account. That would be better for me since I will use Maven to build the modules.

springfox-spring-web/src/main/java/springfox/documentation/spring/web/DocletGenerator.java
```diff
@@ -0,0 +1,279 @@
...
+
+import java.io.File;
+import java.io.FileOutputStream;
+import java.io.OutputStream;
+import java.util.Properties;
+
+/**
+ * Generate properties file based on Javadoc.
+ */
+public class DocletGenerator  extends Doclet {
```
rgoers: I can do that if I am using Maven. I have no clue how your tests run.

springfox-spring-web/src/main/java/springfox/documentation/spring/web/plugins/JavaDocBuilderPlugin.java
```diff
@@ -0,0 +1,158 @@
...
+import springfox.documentation.spi.service.ParameterBuilderPlugin;
+import springfox.documentation.spi.service.contexts.OperationContext;
+import springfox.documentation.spi.service.contexts.ParameterContext;
+
+/**
+ * Swagger JavaDoc properties builder plugin.
+ */
+@Component
+@Order(Ordered.LOWEST_PRECEDENCE)
+public class JavaDocBuilderPlugin implements OperationBuilderPlugin, ParameterBuilderPlugin {
```
rgoers: That is correct. It would be nice to have the return information but I am not sure where to inject it. I suspect throws makes no sense since a REST service shouldn't throw anything.

springfox-spring-web/src/main/java/springfox/documentation/spring/web/plugins/JavaDocBuilderPlugin.java
```diff
@@ -0,0 +1,158 @@
...
+
+        String notes = context.requestMappingPattern() + PERIOD +
+                context.httpMethod().toString() + ".notes";
+        if (StringUtils.hasText(notes) && StringUtils.hasText(propertiesReader.getProperty(notes))) {
+            context.operationBuilder().notes(propertiesReader.getProperty(notes));
+        }
+    }
+
+    @Override
+    public void apply(ParameterContext context) {
```
rgoers: I used reflection because there isn't a dependency on io.swagger and I ddin't want to add that dependency. I'm not sure why you don't have it.

springfox-spring-web/build.gradle
dilipkrish: I think that might be the better option. Plus you can control your timelines for release to suite your dependent teams. Once you think its ready we can and port it to the `springfox` organization as a separate library, that is of course, if you'd like to contribute it back for others to use. Would love to be a add-on library like the [grails integration](https://github.com/springfox/springfox-grails-integration)

springfox-spring-web/src/main/java/springfox/documentation/spring/web/DocletGenerator.java
dilipkrish: Maven/Gradle both support regular junit tests, [just like this one](https://github.com/springfox/springfox/blob/master/springfox-spring-web/src/test/java/springfox/documentation/spring/web/ObjectMapperConfigurerIntegrationTest.java).

springfox-spring-web/src/main/java/springfox/documentation/spring/web/plugins/JavaDocBuilderPlugin.java
dilipkrish: That is true, but throws is can describe what we can add to the operation notes. e.g. Free text something like this => `IllegalArgumentsException when invalid input translates to a 400`

springfox-spring-web/src/main/java/springfox/documentation/spring/web/plugins/JavaDocBuilderPlugin.java
dilipkrish: Hmm this reveals an interesting gap. Really in this case you should have an API provided by springfox like `ParameterContext#updateProperties(Environment env)` So the context can load the properties it cares about like description etc. Remember it already knows its name/isrequired etc from other plugins reading the spring annotations :)

springfox-spring-web/build.gradle
rgoers: Is it ok if I call the project springfox-javadoc and use the springfox package names as it currently is doing? I would prefer to give this back to you as I won't be able to publish it to Maven Central using your groupId.

springfox-spring-web/src/main/java/springfox/documentation/spring/web/plugins/JavaDocBuilderPlugin.java
rgoers: OK - but where do I inject that so it prints?

dilipkrish: In the operation notes/description?

springfox-spring-web/build.gradle
dilipkrish: That sounds perfect

andye2004:
Been thinking about this for a little while and I think I would take a slightly different approach. I'm thinking it would be possible to use a maven javadoc plugin with a javadoc to json doclet producer and then use jsonpath to read in the javadoc at runtime. Might be interested in working on this as a plugin in my spare time. Do you guys have anything, guides, documentation etc on how to create a plugin or where the plugin hooks are and how to use them?

andye2004:
Having read over the work done by @rgoers seems silly to replicate the work.

@dilipkrish Any indication when this will be brought in and released?

rgoers:
@andye2004 To be honest I've completely put this on the back burner. I've never liked that springfox has to execute at run time. We are currently looking into the maven-swagger-plugin as an alternative.

dilipkrish:
@rgoers have you considered generating swagger specification @ build time using an integration test as suggested by [swagger2markup](https://github.com/Swagger2Markup/swagger2markup)? Its definitely higher fidelity of models than the maven plugin.

dilipkrish:
@andye2004 I'd love to bring this in. Just need more time to prioritize this

MartinNeumannBeTSE:
Hi, based on this code, I have created my own version of the doclet and plugin, which include some fixes and enhancements. I'd like to contribute those as a separate module, like @dilipkrish suggested. Can you help me set this up?

dilipkrish:
@MartinNeumannBeTSE Absolutely. What do you need help with?

MartinNeumannBeTSE:
@dilipkrish Maybe we should switch to private communication because instructions on how to organize my contribution are not really part of the pull request.
I basically only have two classes: the doclet and the Springfox plugin. I would also write a readme and probably some tests. They should probably be part of a separate module because maybe not everybody using Springfox wants to have this activated.
How would I set up such a module? Do I create my own repository for that, or do I fork the Springfox repository and put the classes, tests, and documentation there? If so - where?

springfox-spring-web/src/main/java/springfox/documentation/spring/web/plugins/JavaDocBuilderPlugin.java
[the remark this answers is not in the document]
MartinNeumannBeTSE: In my version I already have the `@return` tag mapped (to the summary) and I've started mapping the `@throws` tags, but I'm also struggling where to put this information. This could probably be integrated with an implementation for #521

dilipkrish:
@MartinNeumannBeTSE here you go -> https://github.com/springfox/springfox-javadoc

dilipkrish:
Closing this PR since this is superseded by https://github.com/springfox/springfox-javadoc

rgoers:
It is unfortunate that in copying the code any record of my commits were lost.

dilipkrish:
Sorry about that! Wasn’t aware it was copied without attribution. Thank you for the excellent idea. While I know it’s not the same, I’ll happily add a note in the readme. Additionally, would you like me to add you as a collaborator?

rgoers:
Thanks, but a note isn't necessary. Yes, you can add me as a collaborator.

dilipkrish:
@rgoers I've added you as a collaborator.

MartinNeumannBeTSE:
@rgoers I did not intend in any way to take credit for your work. If @dilipkrish agrees, we could add @author javadoc tags to the classes, so that you get the credit you deserve.

dilipkrish:
@MartinNeumannBeTSE my fault for not thinking about that, sorry. A better place to add IMO is the pom file developers section, but adding it as an `@author` tag some how seems appropriate for _this_ project :)

Please also don't discount your work in anyways in taking the idea forward.

MartinNeumannBeTSE:
I added a developers section to the POM file

MartinNeumannBeTSE:
I also added the `@author` tags to the class-level javadoc

Issue #1818, opened 2017-05-16
JsonFormat for enum and other cases


yelhouti:
Please take the time to search the repository, if your question has already been asked or answered.

- [2.6.1 latest] What version of the library are you using? Is it the latest version? The latest released version is [ ![Download](https://api.bintray.com/packages/springfox/maven-repo/springfox/images/download.svg) ](https://bintray.com/springfox/maven-repo/springfox/_latestVersion)

What kind of issue is this?

 - [x] Bug report. If you’ve found a bug, spend the time to write a failing test. Bugs with tests or 
       steps to reproduce get fixed faster. Here’s an example: https://gist.github.com/swankjesse/6608b4713ad80988cdc9
   
   `@JsonFormat(shape=JsonFormat.Shape.OBJECT)` is not supported
   this property allows to threat enums as objects:
  
example: 

```
@JsonFormat(shape= JsonFormat.Shape.OBJECT)
@ApiModel(value = "Card Error", description = "Information about the error")
public enum CardError {
    BLOCKED("Token Blocked"),
    EXPIRED("Token Expired"),
    UNKNOWN("Token Unknown");
    @ApiModelProperty(value = "Error description")
    private String description;

    CardError(String description) {
        this.description = description;
    }

    public String getDescription() {
        return description;
    }

    public String getValue(){
        return this.name();
    }
}
```

Controller Methode:
```
    @RequestMapping(value = "/error",method = RequestMethod.GET)
    @ApiOperation(value = "get Error", notes = "test get Error", response = CardError.class)
    @ApiResponses({
            @ApiResponse(code = 401, message = "BLOCKED", response = CardError.class)
    })
    public ResponseEntity<CardError> getError(){
        return new ResponseEntity<>(CardError.EXPIRED, HttpStatus.UNAUTHORIZED);
    }
```

What is returned by spring :
```
{
  "description": "Token Expired",
  "value": "EXPIRED"
}
```
What Swagger displays:
```
HTTP Status Code   Reason        Response Model    Headers
401                BLOCKED       string
```

what happens is that Swagger doesn't take into account the JsonFormat and instead convert the enum to a String.

What do you think please

yelhouti:
After some digging in the code this what I have understood so fare:
The problem happens in `springfox/springfox-core/src/main/java/springfox/documentation/schema/Enums.java` when getting allowable values.
[...]

PR #1829, opened 2017-05-22
JacksonEnumDeterminer to handle JsonFormat.Shape.Object

yelhouti:
#### What's this PR do/fix?
Handles JsonFormat.Shape.Object when used on Enums properly. to Threat them as objects instead of Enums
Yes
Mabye it's a good idea to provide integration tests
https://github.com/springfox/springfox/issues/1818


yelhouti:
Could you please re run the test, jcenter.bintray.com was donw for few seconds I think

dilipkrish:
Are you trying to compile it with gradle before you pushed it up? None of the tests are compiling

dilipkrish:
Also would you mind adding a contract/integration test for your use case? There seems to be no tests for that. Basically it would exercise your change in the [`FeatureDemostrationService`](https://github.com/springfox/springfox/blob/master/springfox-spring-web/src/test/java/springfox/documentation/spring/web/dummy/controllers/FeatureDemonstrationService.java) and will add to the demonstrated features.

dilipkrish:
Thanks @yelhouti

yelhouti:
Thank you :)

Issue #1903, opened 2017-07-07
XmlAttribute not applied


janfockaert:
Latest Swagger2 version 2.7.0

We have a Spring Boot application that exposes XML api and we want to show a correct XML example in the Swagger spec. The @XmlAttribute it is not applied on the spec but shown as an XML element. I already searched for other issues and added the jackson jaxb module.

On the following [sample repo](https://github.com/merapar/spring-boot-swagger-sample) you can reproduce this behaviour.

The expected outcome in Swagger Example model
```xml
<Sample code="1" value="string">
</Sample>
```
but is:
```xml
<Sample>
	<code>1</code>
	<value>string</value>
</Sample>
```

Some code samples copied from the sample repo
Application.java:
```java
@SpringBootApplication
@EnableSwagger2
public class Application implements ApplicationListener<ObjectMapperConfigured> {
	public static void main(String[] args) throws Exception {
		SpringApplication.run(Application.class, args);
	}

	@Bean
	public Docket api() {
		return new Docket(DocumentationType.SWAGGER_2)
				.select()
				.apis(RequestHandlerSelectors.basePackage("com.merapar.controller"))
				.paths(PathSelectors.ant("/*"))
				.build();
	}

	@Override
	public void onApplicationEvent(ObjectMapperConfigured event) {
		event.getObjectMapper().registerModule(new JaxbAnnotationModule());
	}
}
```
Sample.java:
```java
@NoArgsConstructor
@AllArgsConstructor
@XmlRootElement(name = "Sample")
public class Sample {

    @Getter
    @XmlAttribute(name = "value", required = true)
    private String value;

    @Getter
    @XmlAttribute(name = "code")
    private CodeEnum code;

    @XmlEnum
    public enum CodeEnum {
        @XmlEnumValue("1")
        ENUM1(1),
        @XmlEnumValue("2")
        ENUM2(2);

        private int value;

        CodeEnum(int value) {
            this.value = value;
        }

        @Override
        public String toString() {
            return String.valueOf(value);
        }
    }
}
```

Not sure whether this is a bug or a feature request.

Thanks in advance.

dilipkrish:
I suspect it isn't respecting the model in swagger-ui correctly. Not sure what to do about that.

janfockaert:
It's missing the following attributes on the properties with @XmlAttribute:
```json
  "xml": {
    "attribute": true
  }
``` 

so the complete Swagger2 should be:
```json
{
    "swagger": "2.0",
[...]

Issue #1902, opened 2017-07-07
XML example value has incorrect structure


display-none:
I have an issue with springfox swagger 2.7.0.

As a response body I have the following:

```java
public final class Response {
	@JacksonXmlElementWrapper(localName = "partialResults")
	@JacksonXmlProperty(localName = "partialResult")
	private final Set<PartialResult> partialResults;
	private final Status finalResult;
	private final String version;

	 ...
}
```

```java
public final class PartialResult {
	private final String name;
	private final Status result;
	@JsonInclude(Include.NON_EMPTY)
	private final String reason;

	 ...
}
```

This call yields for example following response:

```xml
<Response>
  <finalResult>NOK</finalResult>
  <version>3.15</version>
  <partialResults>
    <partialResult>
      <name>rule1</name>
      <result>OK</result>
    </partialResult>
    <partialResult>
      <name>rule2</name>
      <result>NOK</result>
      <reason>something failed</reason>
    </partialResult>
  </partialResults>
</Response>
```

In swagger ui I can see the model is correct:
```
Response {
finalResult (string, optional) = ['NOK', 'OK', 'UNKNOWN'],
partialResults (Array[PartialResult], optional),
version (string, optional)
}
PartialResult {
name (string, optional),
reason (string, optional),
result (string, optional) = ['NOK', 'OK', 'UNKNOWN']
}
```

But example value is not:
```xml
<?xml version="1.0"?>
<Response>
  <finalResult>NOK</finalResult>
  <partialResults>
    <name>string</name>
    <reason>string</reason>
    <result>NOK</result>
  </partialResults>
  <version>string</version>
</Response>
```

Notice that under `partialResults` there is always a `partialResult` tag, which is missing in example value.

dilipkrish:
Please report this issue in swagger-ui

PR #1907, opened 2017-07-08
Xml

nickus:
#### What's this PR do/fix?
This PR adds XML supporting for models. 
https://github.com/OAI/OpenAPI-Specification/blob/master/versions/2.0.md#xmlObject


dilipkrish:
@nickus thanks for the PR. Not to get stuck on coverage numbers alone, but it would be useful to add some tests.

dilipkrish:
This is related to #1902 and #1903

nickus:
OK

nickus:
done

springfox-schema/src/main/java/springfox/documentation/schema/DefaultModelProvider.java
```diff
@@ -109,10 +109,20 @@ private Model modelBuilder(ResolvedType propertiesHost,
...
     return schemaPluginsManager.model(modelContext);
   }
 
+  private Xml buildXml() {
+    return new Xml()
+        .name("")
+        .attribute(false)
+        .namespace("")
+        .prefix("")
+        .wrapped(false);
```
dilipkrish: This doesnt seem right. We shouldn't always initialize the xml. this has resulted in all the contracts having the xml element in them.

Also we need a way to grab the right xml namespace/prefix/name from the `XmlSchema` annotation.

springfox-core/src/main/java/springfox/documentation/builders/ModelPropertyBuilder.java
```diff
@@ -103,6 +105,12 @@ public ModelPropertyBuilder extensions(List<VendorExtension> extensions) {
     return this;
   }
 
+  public ModelPropertyBuilder xml(Xml xml) {
+    this.xml = defaultIfAbsent(xml, this.xml);
+    return this;
+  }
```
dilipkrish: This method is ever used. Was that intentional?

dilipkrish:
@nickus I polished and merged your PR. Yet to be done for complete support is reading the XmlSchema by package and adding support fo xmlns prefixes. Would love it if you could complete the work for Xml support.

dilipkrish:
I created an XmlModelPlugin

https://github.com/springfox/springfox/blob/640c75a6900fa284c754e7dc4cc4c3982e7f4469/springfox-schema/src/main/java/springfox/documentation/schema/XmlModelPlugin.java#L42-L45

and an XmlPropertyPlugin

https://github.com/springfox/springfox/blob/master/springfox-schema/src/main/java/springfox/documentation/schema/property/XmlPropertyPlugin.java#L43-L82

Would be awesome if you verified the desired output as well

jgaribay21:
looks like there is a missing check optional check here:
`  Optional<XmlElementWrapper> wrapper = findPropertyAnnotation( 
       context.getBeanPropertyDefinition().get(), 
       XmlElementWrapper.class); `

If it is a field that has the XmlElement annotation this will cause an IllegalStateException.  It should be checking if it is a bean, and if not if the wrapper is annotated on the field:

`@XmlElementWrapper(name = "TestElements")
@XmlElement(name = "TestElement")
protected List<SomeObject> testElements = new ArrayList<SomeObject>();`

jgaribay21:
Also, XML properties are lost when running through ApiModelReader.mergeModelMap.  The merge should copy over the Xml node to the new modelbuilder.

dilipkrish:
@jgaribay21 thanks for the feedback! this is helpful. Would you mind creating an issue for it?

PR #1917, opened 2017-07-13
Fix for issue 1916: Consumes / Produces media-types on the document level aren't copied and merged anymore with the operation level consumes / produces media-types.

mzeijen:
This is a possible fix for issue 1916. It removes the merging of the document level and operation level consumes and produces media types on the in the OperationContext. This results in a "more" valid swagger specification because now the document level media types are only defined on the document level and only when these media-types are overwritten on the operation level, they will be set on the operation level. This behaviour is according to the Swagger v2 specification and it also works just fine within the Swagger UI.

However we do want to provide some default values when no media-types are defined on the document and operation level. Thus I changed the MediaTypeReader to reflect this.

I also fixed the contract tests, because many of these didn't produce the correct specification anymore. TBH: I didn't check every specification by hand because many many failed due to this change. For most of them I just made sure that the expected specification matched the produces specification.

### DIFF
--- a/springfox-spi/src/main/java/springfox/documentation/spi/service/contexts/OperationContext.java
+++ b/springfox-spi/src/main/java/springfox/documentation/spi/service/contexts/OperationContext.java
@@ -24,7 +24,6 @@ import com.google.common.base.Optional;
 import com.google.common.base.Predicate;
 import com.google.common.collect.ImmutableSet;
 import com.google.common.collect.Iterables;
-import com.google.common.collect.Sets;
 import org.springframework.http.HttpMethod;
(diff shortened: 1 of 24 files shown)


dilipkrish:
Could you perhaps add some tests for this change?

mzeijen:
No problem, I can add some additional tests.

mzeijen:
@dilipkrish I updated the MediaTypeReaderSpec with additional tests and added two contract tests that test the behaviour for this change. I hope this suffices.

dilipkrish:
:metal: Will pull it in over the weekend. Thanks for being diligent! 🙇

PR #1942, opened 2017-07-26
Model classes having names containing integers are not detected as array

mathieuales:
swagger 1.2 : model classes having names containing integers are not detected as array


dilipkrish:
thanks @mathieuales!

Issue #1880, opened 2017-06-22
wrong API resource path


mathieuales:
Hi !
I'm using `springfox-swagger1:2.6.1` with `springboot 1.5.3`
It seems that the `resourcePath` in the generated swagger.json of a controller having only one operation is not the one defined using annotations

```java
@RestController
@RequestMapping(value = "/supervision", produces = {"application/json"})
@Api(description = "supervision endpoint")
public class Supervision {

        @GetMapping("/status")
        @ApiOperation(value = "global API status")
        public ResponseEntity getStatus() {
              return ....
        }
}
```
The generated JSON for this controller is
```js
{
      "swaggerVersion": "1.2",
      "apis": [{
            ...
            "operations": [{
                  "method": "GET",
                  ...
            }],
            "path": "/supervision/status"
      }],
      "resourcePath": "/supervision/status"
}
```
`resourcePath` should be `"/supervision"` as defined in `@RequestMapping` annotation

Moving to Swagger2 is not an option for now as I'm working on an old API always returning HTTP 200 OK with body containing success or error responses

mathieuales:
Hi !
According to me it looks to be an issue:
if you have a look at `ApiListingScanner::scan` in `springfox-spring-web` the `resourcePath` is determined using function `longestCommonPath`. 
This should be done like this only if annotation `@RequestMapping` of the controller has no value (ie a path)

dilipkrish:
I agree with you, but could you explain how you're using it? It used to work as expected in swagger-ui.

mathieuales:
I'm using my own component to display swagger documentation, and i'm using the resourcePath to group resources defined in different controllers as tags are not usable in swagger 1.2.
I tried [swagger-ui](http://petstore.swagger.io) but it says "No operations defined in spec!"

dilipkrish:
Weird looks like version 1.2 isn't actively supported. Is `angular-swagger-ui` expecting a certain way the resource path is laid out?

mathieuales:
No, there is no restriction on resource path. The issue is just springfox does not set the resource path defined in `@RequestMapping`
Thanks for your time

PR #1943, opened 2017-07-26
Fix issue "wrong API resource path" #1880

mathieuales:
Fix issue "wrong API resource path" #1880


dilipkrish:
How about adding a test for this change

mathieuales:
sorry for this dirty PR, i just can't get all tests properly running even without my modifications

dilipkrish:
@mathieuales thanks again for your contribution 🙇

PR #1956, opened 2017-07-30
fixes #1941

### LINKED ISSUE #1941: swagger-ui does not remove java string escape char "\" from @ApiModelProperty example
Using version 2.7.0 of springfox-swagger2 and springfox-swagger-ui with spring-boot 1.5.4.RELEASE

I have defined some example value in `@ApiModelProperty` like this
```java
@ApiModelProperty(example = "{\"name\": \"foo\", \"age\": 42}")
```
as part of a model for a body param.

The escape chars `\` are kept in swagger-ui's `Example Value`, so that when clicking on it to paste the `Example Value` into body param value for testing the REST endpoint, the escape chars `\` are still included - thus transmitting invalid JSON to the endpoint.

heapifyman:
fixes #1941

### DIFF
--- a/springfox-swagger2/src/main/java/springfox/documentation/swagger2/configuration/Swagger2JacksonModule.java
+++ b/springfox-swagger2/src/main/java/springfox/documentation/swagger2/configuration/Swagger2JacksonModule.java
@@ -123,8 +123,8 @@ public class Swagger2JacksonModule extends SimpleModule implements JacksonModule
         // and an example value of "null" probably does not make much sense anyway
         return value.startsWith("{")                              // object
                 || value.startsWith("[")                          // array
-                || value.equals("true")                           // true
-                || value.equals("false")                          // false
+                || "true".equals(value)                           // true
+                || "false".equals(value)                          // false
                 || JSON_NUMBER_PATTERN.matcher(value).matches();  // number
       }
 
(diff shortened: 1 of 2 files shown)


springfox-swagger2/src/main/java/springfox/documentation/swagger2/configuration/Swagger2JacksonModule.java
```diff
@@ -79,4 +86,64 @@ public void setupModule(SetupContext context) {
...
+
+      private boolean isNotJsonString(final String value) throws IOException {
+        // strictly speaking, should also test for equals("null") since {"example": null} would be valid JSON
+        // but swagger2 does not support null values
+        // and an example value of "null" probably does not make much sense anyway
+        return value.startsWith("{")                              // object
+                || value.startsWith("[")                          // array
+                || "true".equals(value)                           // true
+                || "false".equals(value)                          // false
+                || JSON_NUMBER_PATTERN.matcher(value).matches();  // number
```
dilipkrish: Can we add a serialization json test that covers these criteria?

heapifyman: Yes, sure. But I'm going to need a little help with that.
I am not that familiar with groovy and the spock framework.
Also, where to put the tests? Can I extend some existing test class or create a new one?
Would be great if you could give me some starting point.

dilipkrish: If you find it easier, the standard junit tests work just as well.

PR #1988, opened 2017-08-18
custom the web page title as swagger.title when had set it

rainplus:
custom the web page title as swagger.title when had set it

#### What's this PR do/fix?
When I am developing, I will open a lot of interface documents page, but I can not for which of them which is, only one page to open, so it will modify the code, easy to view from the browser tabs title


dilipkrish:
@rainplus thank you!! 🙇

Issue #1738, opened 2017-03-24
Duplicated enumerations in swagger definitions


Sineaggi:
Version 2.6.1

Enums are currently being listed inline in the swagger output, causing enums to be generated as inner class. They should be referenced via `"$ref"` like classes are. Attached are two swagger json definitions, one provided by springfox and the other fixed by hand.

This is problematic for clients that attempt to use the model files, as they cannot be reused across api endpoints, as each generated model will contain a copy of the enum. This is especially problematic in apis that define enums with large sets of values, such as an enum set of the 249 ISO 3166 alpha 2 or 3 letter country codes.

Example enum:
`public enum Crest {
    ONE,
    TWO,
    THREE,
}`

Example class:
```java
public class Obj {
    private Crest crest;

    public Crest getCrest() {
        return crest;
    }

    public void setCrest(Crest crest) {
        this.crest = crest;
    }
}
```

Incorrect swagger:
```json
"definitions": {
  "Obj": {
    "type": "object",
    "properties": {
      "crest": {
        "type": "string",
        "enum": [
          "ONE",
          "TWO",
          "THREE"
        ]
      }
    }
  },
}
```
Resulting codegen:
```java
// ... snip ...
public class Obj {
  /**
   * Gets or Sets crest
   */
  public enum CrestEnum {
    ONE("ONE"),
    
    TWO("TWO"),
    
    THREE("THREE");

    private String value;

    CrestEnum(String value) {
      this.value = value;
    }
  }
  // ... snip ...
}
```

Correct swagger:
```json
"definitions": {
  "Obj": {
    "type": "object",
    "properties": {
      "crest": {
        "$ref": "#/definitions/Crest"
      }
    }
  },
  "Crest": {
    "type": "string",
    "enum": [
      "ONE",
      "TWO",
      "THREE"
    ]
  },
}
```
Resulting codegen:
```java
// ... snip ...
public enum Crest {
  
  ONE("ONE"),
  
  TWO("TWO"),
  
  THREE("THREE");

  private String value;

  Crest(String value) {
    this.value = value;
  }
  // ... snip ...
}

// ... snip ...
public class Obj {
  @JsonProperty("crest")
  private Crest crest = null;

  // ... snip ...
  public Crest getCrest() {
    return crest;
  }

  public void setCrest(Crest crest) {
    this.crest = crest;
  }
  // ... snip ...
}
```

dilipkrish:
[...]

PR #1991, opened 2017-08-20
Feature/open classes for extension for enum ref

atsu85:
#### What's this PR do/fix?
It changed some modifiers of 4 classes from private to protected (and extracted one method), so enum type references (as proposed with #1738) could be implemented.

I added unit tests with separate commit - testing using reflection that methods have now `protected` modifier, so subclasses can override or invoke them.

Issue: #1738

### DIFF
--- a/springfox-schema/src/main/java/springfox/documentation/schema/DefaultModelDependencyProvider.java
+++ b/springfox-schema/src/main/java/springfox/documentation/schema/DefaultModelDependencyProvider.java
@@ -129,7 +129,7 @@ public class DefaultModelDependencyProvider implements ModelDependencyProvider {
     return parameters;
   }
 
-  private List<ResolvedType> resolvedPropertiesAndFields(ModelContext modelContext, ResolvedType resolvedType) {
+  protected List<ResolvedType> resolvedPropertiesAndFields(ModelContext modelContext, ResolvedType resolvedType) {
     if (modelContext.hasSeenBefore(resolvedType) || resolvedType.getErasedType().isEnum()) {
       return newArrayList();
     }
--- a/springfox-schema/src/main/java/springfox/documentation/schema/DefaultModelProvider.java
+++ b/springfox-schema/src/main/java/springfox/documentation/schema/DefaultModelProvider.java
@@ -72,11 +72,7 @@ public class DefaultModelProvider implements ModelProvider {
   @Override
   public com.google.common.base.Optional<Model> modelFor(ModelContext modelContext) {
     ResolvedType propertiesHost = modelContext.alternateFor(modelContext.resolvedType(resolver));
-    if (isContainerType(propertiesHost)
-        || isMapType(propertiesHost)
-        || propertiesHost.getErasedType().isEnum()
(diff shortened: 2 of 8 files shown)


atsu85:
@dilipkrish, 
could You review this tiny PR please?

dilipkrish:
@atsu85 sorry Im only getting to this PR. Is there are reason you want to extend these? Could you provide an use case/details. In general, Im hesitant to provide extensibility as it increases the surface area for support

atsu85:
@dilipkrish ,
thanks for taking time for this PR. As i stated in the comment, the use case is to allow implementing 
> enum type references (as proposed with #1738)

Right now i have just patched these classes in my project (with the same changes), so that i can sub-class them to implement enum references. 

Ideally i'd like to contribute back not just extension points, that would allow other people (such as @Sineaggi) to implement the solution in a similar way, but a complete enum-references feature (that should be disable out-of-the-box for backwards compatibility), but there is one catch related to OpenApi specification:  
While enum references are allowed for example in definition properties (java class fields), they are not allowed in query parameters, but fortunately i'm not using enums in query parameters (we have already discussed it with [this](https://github.com/springfox/springfox/issues/1738#issuecomment-353336176) and following 3 comments). So for now i haven't had the motivation or time to investigate what else should need to be changed, to avoid using enum references in case of query parameters while still referencing enums from type properties.

robertjamesmccullough:
I have reached out about helping with an ultimate solution for this problem, but in the mean time @atsu85 are you willing to share your solution that builds upon this PR?

atsu85:
@robertjamesmccullough,
>  @atsu85 are you willing to share your solution that builds upon this PR?

Yes, i'll try to prepare a sample project that includes classes that take advantage of this PR to create enum references. However i can't do it right now, so don't expect it in a day or two.

I'll have to note, that enum references are not valid in all places, for example query parameters - my solution doesn't solve that, but otherwise it works well in our project

dilipkrish:
Thank you @atsu85... I'll try to get your solution or something that accomplishes something similar in the next release. Sorry this has been languishing

atsu85:
[As i promised](https://github.com/springfox/springfox/pull/1991#issuecomment-363560177), i tried to add sample code for generating enum-refs into [springfox-petstore](https://github.com/springfox/springfox/tree/master/springfox-petstore) but stumbled upon couple of problems
1) There are some test failures that appear to fail only on windows (see #2257 - fixed only couple of them) - would be nice to verify what tests naive enum-ref implementation would break.
2) Didn't figure out how to start [springfox-petstore](https://github.com/springfox/springfox/tree/master/springfox-petstore) outside tests - `gradlew bootRun -p springfox-petstore` doesn't work.
I'm going to deal with it later, right now i'm going to sleep ;)

dilipkrish:
@atsu85 Trying to validate this PR would you mind making the changes that build on this PR in the [the demos](https://github.com/springfox/springfox-demos) project. You should have instructions to start the project/run it etc. 

Also when u have a chance you need to resolve the conflicts in this PR and squash the merge commits.

Issue #1900, opened 2017-07-05
Tag custom ordering


rainoko:
Right not tags are ordered by name comparator. May be its more useful if tag contains order numer. And comparator should order by that number falling back to name if order by number is equal.

I can contribute with that.

dilipkrish:
Could you describe how you'd provide the order on the tag?

rainoko:
I would add int order for tag. which should be optional and default 0 or max int.  and rewrite namecomparator for using this new field.

public class Tag {
  private final String name;
  private final String description;
  **private final int order;**

https://github.com/rainoko/springfox/commit/bc5af62e9757afb60fdc90a7d8c6a46bf7588e58

dilipkrish:
🎱 Not really looking how to do it. Im asking how you'd specify a value for the order (via annotations etc. I presume?) You'd have to create new annotation because the existing swagger annotations don't support it.

rainoko:
First solution that I thought was through docket.
.tags(new Tag("Pet Service", "All apis relating to pets", 1)) 

So custom ordering supported only when explicitly defining tags in docket.

dilipkrish:
That seems reasonable. How about implementing the spring `Ordered` interface. Happy to take in the contribution.

PR #2048, opened 2017-09-22
Support explicit ordering for Tags

jroweboy:
Adds support for explicitly ordering `Tag`s when manually adding tags to a `Docket`

This code addresses issue https://github.com/springfox/springfox/issues/1900 and is based heavily off rainoko's code that was linked in the issue. A unit test was added to show that ordering tags should be ordered first based on priority and second by name. In order to support a second sorting order, the comparator is kept around, even though tags implement `Ordered` now.

(In other news, I don't know if this has been reported, but running the gradle tests on windows doesn't work if you have configured git to pull with windows line endings and push with unix line endings. This is  because the test cases often compare multiline strings and they will fail because of the added `\r` thanks to windows line endings.)


dilipkrish:
@jroweboy Thanks this is awesome!

PR #1671, opened 2017-02-10
General functionality for comparing models. Initial support for #182, #807, #895, #1356

MaksimOrlov:
Good evening, @dilipkrish .

It is my view and my suggestion for how **models** should be compared and linked with each other.

This PR is related to issues #182, #563, #807, #895 and #1356.

The short review about idea is:

1. At first we add link between **ResolvedMethodParameter** and it's **ModelContext**. We will keep links inside **OperationModelContextsBuilder**.

2. Then, when we will go through each **ModelContext** we will use _.fromParent_ to keep link inside **ModelContext**, because, if we change it's model (_@JsonView_ or something else) we will change parent's model consequently. 

3. Each **ModelProperty** keeps link to related **ModelContext** inside **ModelReferenceProvider**.

4. We compare models with .equals, by building a tree with contexts and compare it with other models level by level.

Comparing models without saving links could lead to the situation when model's name was changed (i.e. _name ->name_1_), but **ModelReference** (in **Parameter** or **ModelProperty**) is still pointing to old model's name, that will lead to inappropriate stuff....

I'm looking forward for your opinion, remarks, suggestions, etc....

P.S I removed **Cacheable** while I was testing, I think I know the way how to return it back.

PR #2056, opened 2017-10-01
PR to support Model enhancements 

dilipkrish:
#Model enhancements include
- [x] Support for JsonView

## Part of a different effort (these are different concepts) edited `2017-10-07 10:51:59`
- [ ] Support for spring data rest projects
- [ ] Support for inheritance
- [ ] Possibly support all of these in the light of OAS 3.0


springfox-schema/src/main/java/springfox/documentation/schema/ModelProjectionExtractor.java
```diff
@@ -0,0 +1,63 @@
...
+      @Qualifier("modelProjectionProviderRegistry")
+      PluginRegistry<ModelProjectionProviderPlugin, DocumentationType> modelProjectionProviders) {
+
+    this.typeResolver = typeResolver;
+    this.modelProjectionProviders = modelProjectionProviders;
+  }
+  
+  public Optional<ResolvedType> extractProjection(ResolvedType type, List<Annotation> annotations, DocumentationType documentationType) {
+    ModelProjectionProviderPlugin selected =
+            modelProjectionProviders.getPluginFor(documentationType, new JacksonJsonViewProjectionProvider());
```
dilipkrish: I think if there are plugin extensibility's we should use all of them not just the first one. That way we can layer information via plugins.

springfox-schema/src/main/java/springfox/documentation/schema/JacksonJsonViewProjectionProvider.java
```diff
@@ -0,0 +1,29 @@
...
+
+public class JacksonJsonViewProjectionProvider implements ModelProjectionProviderPlugin {
+
+  @Override
+  public boolean supports(DocumentationType delimiter) {
+    return true;
+  }
+
+  @Override
+  public Optional<Class<?>> projectionFor(ResolvedType type, List<Annotation> typeAnnotations) {
```
dilipkrish: This should return a `ResolvedType` IMO.

Also have you thought about how we might influence a projection not using annotations (I don't know how gson handles this)? I think it might be a good idea to extract a parameter object i.e. projection context that gives it access to/or parts of the `RequestMappingContext`

dilipkrish: Oh never mind... I see what this is doing. I don't think this interface is accurate. I'm thinking of the return value as an __alternate type__ for the `type`. This makes this interface is jackson specific and will not work for spring-data-rest I think. 

I was thinking more in these terms. Lets take this example [from here](http://www.baeldung.com/jackson-json-view-annotation)

```java
public class Views {
    public static class Public {
    }
 
    public static class Internal extends Public {
    }
}

public class Item {
  
    @JsonView(Views.Public.class)
    public int id;
 
    @JsonView(Views.Public.class)
    public String itemName;
 
    @JsonView(Views.Internal.class)
    public String ownerName;
}
```

I would think that this interface would behave something like this.

```java
//Pseudo code on how this should be used.

ProjectionContext context = new ProjectionContext(
          type, //Type to find a projection type
          annotations, //Annotations to help finding projections
          ... more params as needed
         );
JacksonJsonViewProjectionProvider plugin = createInstanceOfPlugin(); 
Optional<ResolvedType> projection = instance.projectionFor(context);
//OR
Optional<Model> projection = instance.projectionFor(context);
```
Here projection would be a either a in-memory `ResolvedType` of type `ItemPublicView` built using [AlternateTypeBuilder](https://github.com/springfox/springfox/blob/master/springfox-core/src/main/java/springfox/documentation/builders/AlternateTypeBuilder.java ) or straight away building the Model object via the `ModelBuilder`, that has the public properties as we do here. What do you think?

springfox-spi/src/main/java/springfox/documentation/spi/service/contexts/RequestMappingContext.java
```diff
@@ -187,6 +187,10 @@ public String getGroupName() {
   public <T extends Annotation> Optional<T> findControllerAnnotation(Class<T> annotation) {
     return handler.findControllerAnnotation(annotation);
   }
+  
+  public List<Annotation> getAnnotations() {
```
dilipkrish: Does the `findControllerAnnotation` `findAnnotations` not work? I feel like a plugin should request only annotations it knows about. This is very open interface in my opinion.

springfox-spring-web/src/main/java/springfox/documentation/spring/web/WebMvcRequestHandler.java
```diff
@@ -114,6 +116,11 @@ public String getName() {
   }
 
   @Override
+  public List<Annotation> getAnnotations() {
```
dilipkrish: Same here... I feel like a plugin should request only annotations it knows about. This is very open interface in my opinion.

springfox-spring-web/src/main/java/springfox/documentation/spring/web/readers/operation/.gitignore
```diff
@@ -0,0 +1 @@
+/OperationResponseClassReader.java
```
dilipkrish: ???

springfox-spi/src/main/java/springfox/documentation/spi/service/ProjectionProviderPlugin.java
```diff
@@ -0,0 +1,56 @@
...
+import java.util.List;
+
+import org.springframework.plugin.core.Plugin;
+
+import com.fasterxml.classmate.ResolvedType;
+import com.google.common.base.Optional;
+
+import springfox.documentation.spi.DocumentationType;
+
+public interface ProjectionProviderPlugin extends Plugin<DocumentationType> {
```
dilipkrish: I feel like this plugin shouldn have something other than`DocumentationType` as the discriminator. It should be something like `ProjectionType` which can have one of the following types
- Jackson
- Spring-Data-Rest
- Something we haven't thought off.

Then we can select the right plugin rather than pick the first one the pluginManager finds.Thoughts?

springfox-spi/src/main/java/springfox/documentation/spi/service/ProjectionProviderPlugin.java
```diff
@@ -0,0 +1,56 @@
...
+  List<ResolvedType> projectionsFor(ResolvedType type, Optional<? extends Annotation> requiredAnnotation);
+  
+  /**
+   * Given a type provides is it in a active projection
+   * @param activeProjection - resolved projection type
+   * @param type - type to apply projection
+   * @param annotation - projections of type 
+   * @return resolved projection names
+   */
+  boolean applyProjection(ResolvedType activeProjection, ResolvedType typeToApply, List<ResolvedType> typeProjections);
```
dilipkrish: I did not understand this method. I'll bring up the code and take a look at it again.

springfox-spring-web/src/main/java/springfox/documentation/spring/web/readers/operation/.gitignore
MaksimOrlov: Something wrong, I accidentally ignored.... I will fix it

springfox-spi/src/main/java/springfox/documentation/spi/service/ProjectionProviderPlugin.java
MaksimOrlov: For that we should have some explicit configuration. Any way we won't have all plugins at one time. So, Jackson JsonView Plugin has the lowest priority. If Spring-Data-Rest is included in classpath it will overwrites with its own plugin with higher priority. And custom plugin has the highest priority will overwrite all plugins.

springfox-schema/src/main/java/springfox/documentation/schema/JacksonJsonViewProjectionProvider.java
MaksimOrlov: As i know, Gson does not have full projection support. Just only `SerializationStrategy` to determine which properties will be serialized.

springfox-spi/src/main/java/springfox/documentation/spi/service/contexts/RequestMappingContext.java
MaksimOrlov: I've deleted that already. The main idea is to make common interface to handle all cases with projections.

springfox-spring-web/src/main/java/springfox/documentation/spring/web/WebMvcRequestHandler.java
MaksimOrlov: Agreed, last commit restored that.

springfox-spring-web/src/main/java/springfox/documentation/spring/web/readers/operation/.gitignore
MaksimOrlov: Fixed!

springfox-schema/src/main/java/springfox/documentation/schema/JacksonJsonViewProjectionProvider.java
MaksimOrlov: There is a method `Optional <Class<?> getRequiredAnnotation()` that will handle case with **JsonView**, and will be ignored in case of *Spring Data Rest*

MaksimOrlov: And we might have a case< where there won't be any projection, just all Class... For example, no defined `@Projection` classes, or no **@JsonView** annotation on field.

MaksimOrlov: Dilip, that is outdated code....

```
 @Override
  public List<ResolvedType> projectionsFor(ResolvedType type, Optional<? extends Annotation> requiredAnnotation) {
    List<Class<?>> projections = newArrayList();
    if (requiredAnnotation.isPresent() &&
        requiredAnnotation.get() instanceof JsonView) {
      projections = newArrayList(((JsonView)requiredAnnotation.get()).value());
    }
    return from(projections).transform(toResolvedType()).toList();
  }
```

MaksimOrlov: In last commit I changed interface and implementation

MaksimOrlov: I agreed, `ProjectionContext` is a good idea, to consolidate all required information to determine projection

MaksimOrlov: But, for me *Projection* is **ONLY** a part of `ModelContext`, just additional information, to determine which properties must be serialized or deserialized.

springfox-spi/src/main/java/springfox/documentation/spi/service/ProjectionProviderPlugin.java
MaksimOrlov:
```
 @Override
  public boolean applyProjection(ResolvedType activeProjection, ResolvedType typeToApply,
      Optional<? extends Annotation> requiredAnnotation) {
    final Class<?> activeView = activeProjection.getErasedType();
    if (requiredAnnotation.isPresent() &&
        requiredAnnotation.get() instanceof JsonView &&
        activeView != null) {
      final Class<?>[] typeProjections = ((JsonView)requiredAnnotation.get()).value();
      int i = 0, len = typeProjections.length;
      for (; i < len; ++i) {
        if (typeProjections[i].isAssignableFrom(activeView)) {
          return true;
        }
      }
    }
    return false;
  }
```

MaksimOrlov: Determine if class field belongs to given projection or not

springfox-schema/src/main/java/springfox/documentation/schema/JacksonJsonViewProvider.java
```diff
@@ -0,0 +1,115 @@
...
+  @Override
+  public Optional<ResolvedType> viewFor(ResolvedType type, OperationContext context) {
+    return viewFor(type, context.findAnnotation(JsonView.class));
+  }
+  
+  private Optional<ResolvedType> viewFor(ResolvedType type, Optional<JsonView> annotation) {
+    Optional<ResolvedType> view = Optional.absent();
+    if (annotation.isPresent()) {
+      Class<?>[] views = ((JsonView)(annotation.get())).value();
+      view = Optional.of(typeResolver.resolve(views[0]));
```
artemabalmasov: views[0] might be good for all but ResolvedMethodParameter objects. 
Spring supports single JsonView on endpoint definition, but multiple then put on a field:
```
  @JsonView({View.User.class,View.Practice.class})
  private List<Object> objects;
```
this one should be perfectly fine for a domain object.

MaksimOrlov: That method takes view only from class level.
`applyView` method code works with fields level.

MaksimOrlov: And supports a few views

artemabalmasov: Great, thank you for clarification

jcarnus:
I dont see where the recursion on the given JsonView to get all extendion class or interface is done ?

jcarnus:
the isAssignableFrom ?

MaksimOrlov:
@jcarnus 
Check, please, this [line of code](https://github.com/springfox/springfox/pull/2056/files/685a0d45807d53e05292e9ccaba33645262d542a#diff-840e01ef8971afe9478a4df365702518R102)

artemabalmasov:
@MaksimOrlov  @dilipkrish I know this is silly question, but do you have release date in mind ?

MaksimOrlov:
I'm planning to finish all changes in a next month, stabilization to that PR is almost finished, but I have to integrate [that pull](https://github.com/springfox/springfox/pull/1671) request as well to main line, to make working @JsonView, validation groups, serialization configs, otherwise models will be just merged with each other, losing all difference.

dilipkrish:
@artemabalmasov trying to get it out in the next few weeks.

MaksimOrlov:
Strange behavior of functional tests. There is no errors when I built project under JDK 8, and same two test falls on my laptop under JDK 7 as Circle CI gives . Really curious what is the reason :)

dilipkrish:
I'll look but, it may be because of including Jdk8 types, which tests? I'll look

MaksimOrlov:
The tests are:
should honor api v1.2 contract [declaration-feature-demonstration-service.json]
should honor api v1.2 contract [declaration-pet-service.json]

The main reason is:  Unexpected: Entry_5«string,Pet» 

and there is model Entry_4«string,Pet» two models must be merged into one, but they are appears different....

MaksimOrlov:
fixed. But CI fell with strange error...

tkhalid1:
when the PR is scheduled to be merged into the next release?

dilipkrish:
I think this PR is done. It needs to be rebased with master. correct @MaksimOrlov?

MaksimOrlov:
Yes, @dilipkrish. Should I merge this branch with **master**?

dilipkrish:
I’ll give it a shot. Will give me a chance to take a look at the changes outside of github too. I may message u if I have questions.

MaksimOrlov:
Ok, sure.

swagger-contract-tests/src/test/resources/contract/swagger2/declaration-spring-data-rest.json
```diff
@@ -223,7 +223,7 @@
           "200": {
             "description": "OK",
             "schema": {
-              "$ref": "#/definitions/ResourceOfPerson"
+              "$ref": "#/definitions/Resource_1OfPerson_1"
```
dilipkrish: @MaksimOrlov Im getting to merging this PR finally. 

However, this doesn't look right to me... There shouldn't be 10 different `Resources`. We shouldn't have n different versions of Resource when the model for sure doesn't ever change and is only used in return values.

I would expect `Resource_2OfAddress_1`, `Resource_1OAddress_1` so on and so forth

springfox-swagger2/src/test/groovy/springfox/documentation/swagger2/mappers/ModelMapperSpec.groovy
```diff
@@ -94,17 +102,21 @@ class ModelMapperSpec extends SchemaSpecification {
...
             DocumentationType.SWAGGER_2,
+            typeNameAdjuster,
             alternateTypeProvider(),
             namingStrategy,
             ImmutableSet.builder().build()))
     when:
       def mapped = Mappers.getMapper(ModelMapper).mapModels(modelMap)
     then:
-      mapped.containsKey("MapOfstringAndstring")
+      mapped.containsKey("1484189964")
```
dilipkrish: This seems like it will make things very difficult to troubleshoot. Is there a better way to do this?

swagger-contract-tests/src/test/resources/contract/swagger2/declaration-spring-data-rest.json
MaksimOrlov: @dilipkrish, there is no 10 different `Resources`

So, we have here:

`Resource_1OfPerson_1`
`Resource_2OfAddress_1`
`Resource_3OfCategory_1`
`Resource_4OfTag_1`

`Resources_1OfListOfPerson_1`
`Resources_2OfTag_1`
`Resources_3OfPerson_1`
`Resources_4OfAddress_1`
`Resources_5OfCategory_1`

```
public class Resource<T> extends ResourceSupport {

	private final T content;

        @JsonUnwrapped
	@XmlAnyElement
	public T getContent() {
		return content;
	}
}
```

So, at the runtime T can be any model, and **model** `Resource` has a `ModelRef` inside, that points to any **model**. We compare models at the runtime, SpringFox can see that type of the parameters are different, so models are different. Will name it and increase the digit.

Here we have Person_1 and Person_2, Address_1 and Address_2 and so on, because of ceclic links.

springfox-swagger2/src/test/groovy/springfox/documentation/swagger2/mappers/ModelMapperSpec.groovy
MaksimOrlov: `Id` - is a ```hashCode()``` of the `ModelContext`
`name` - is a normal human readable name of the model.

We need id's to differ models if they have same name.

swagger-contract-tests/src/test/resources/contract/swagger2/declaration-spring-data-rest.json
dilipkrish: Yeah I see that, Im wondering if we can make it better. It will be weird for code generation as well based on these models.

For e.g. `Resources_1OfListOfPerson_1` doesnt have Resource_1**List_1**OfPerson, can we not have Resource also be similar to `List`?

MaksimOrlov: `List` is not a model, it is ignored by `ModelProvider` and `TypeNameExtractor`.

I've added `Map` as an exception, see [that line](https://github.com/springfox/springfox/pull/2056/files#diff-4e40ae816770c5033e017cebee12a6acR141). We can do something like that. F.e. add *NotAdjustableTypes*

But here could be conflicts with user defined types with name *Map* or *Resource*.

jcarnus:
Hi @dilipkrish ,
Is it possible to get a status on this PR and the original one ? 
Thanks

MaksimOrlov:
It is being merged with 2.9.2 tag.

jcarnus:
2.9.2 ? It's already out ?

MaksimOrlov:
Yes, 2.9.2 version was released on the begging of the month. But that PR was not in this version. I'm currently merging it and providing optimization and some other stuff for cyclic links between types.

jcarnus:
Thanks, let me know when the 2.9.2 is updated or 2.9.3 released
Thanks

MaksimOrlov:
ok, sure.

dilipkrish:
@jcarnus Just to clarify...  There is no plan to release 2.9.3. This PR will be part of 3.0.0. /c @MaksimOrlov 

To explain it a little further, this PR has been open for a while, and there are breaking changes in master that this PR does not have. To make it easier/simpler to get this in master sooner, I suggested, not rebasing with 3.0.0-SNAPSHOT(master). Once this PR works with the 2.9.3-SNAPSHOT codebase (2.9.2 tag), I'm happy to resolve the conflicts with master in a separate step and merge it in. That way we isolate this PR and not mix changes from master like java 8 support, spring 5/spring boot 2.0 etc.

Splash34:
Any expected date of releasing this?

MaksimOrlov:
@Splash34, I'm working on some enhancements, and going to finish around beginning of November. If @dilipkrish will approve, then PR will be merged.

jcarnus:
Hi any news on this ?

dilipkrish:
@jcarnus @MaksimOrlov is finishing up with the changes so that it can be pulled in.

leishiyao:
+1

MaksimOrlov:
Hi, @PrabhuKandasamy. 
I'm going to push a new algorithm that can handle cyclic references in a few weeks.

PrabhuKandasamy:
> Hi, @PrabhuKandasamy.
> I'm going to push a new algorithm that can handle cyclic references in a few weeks.

Ok. Waiting for that.

phirzel:
Hey, we would highly appreciate such an improvement as fast as possible.

(Currently we have to deploy each Service-Model resulting in a name clash on separate instances (for e.g. breaking changes, Openshift-Projects), which is not what we really want for operational reasons.)

abdulhafidz:
what happened

MaksimOrlov:
I've pushed changes. Still a few questions remains opened.

dilipkrish:
@MaksimOrlov thank you! thank you! thank you! for working on this! 🙇 

For all of you waiting for this feature, please provide us feedback.

joao-rebelo:
Hi,

  Was having weird bugs since yesterday, untill I realized the snapshot version has changed! (the good and bad of working with snapshots)
  For what I was seeing, it was duplicating my models for each usage in diferent methods. 
  So 3 methods returning ObjectA and I would get model ObjectA, ObjectA_1, ObjectA_2.

  I wasn't able to easily make a simple project to exemplify this.. but if you need I can try.

  Thanks for new developments either way!

Regards

MaksimOrlov:
@joao-rebel, thanks for the reporting the issue.
Could you, please, provide code listing for the  **ObjectA** and for the controllers that use that class?

joao-rebelo:
Hi, 

  Follows the snippet of my code:

RestController:
```
@RestController
@RequestMapping(path = "/basicdata/translation")
@Api(tags = "Translations", description = "Responsible for handling Translations", authorizations = {
		@Authorization("Bearer") })
public class TranslationRestController {

	@Autowired
	private TranslationService translationService;

	@ApiOperation(value = "Bulk get translations", notes = "Bulk returns a set of translations and returns a maps of Xentis IDs to a resolved multi-language translations")
	@GetMapping(path = "/bulk")
	@Nonnull
	public Mono<Map<HexId, MultiLangTranslation>> bulkGet(
			@ApiParam(value = "Ids of translations to fetch", required = true, allowMultiple = true) @RequestParam(name = "id", required = true) @Nonnull Set<HexId> theIds,
			@ApiParam(value = "Types of translations to get", required = true, allowMultiple = true) @RequestParam(name = "type", required = true) @Nonnull Set<TranslationType> theTypes) {
		return translationService.bulkGet(theIds, theTypes);
	}

	@ApiOperation(value = "Get all translations", notes = "Gets all the translations of the specified types and returns a maps of Xentis IDs to a resolved multi-language translations")
	@GetMapping(path = "")
	@Nonnull
	public Mono<Map<HexId, MultiLangTranslation>> getAll(
			@ApiParam(value = "Types of translations to get", required = true, allowMultiple = true) @RequestParam(name = "type", required = true) @Nonnull Set<TranslationType> theTypes) {
		return translationService.getAll(theTypes);
	}

	@ApiOperation(value = "Get all translations in a range", notes = "Gets all the translations within a specific range and returns a maps of Xentis IDs to a resolved multi-language translations")
	@GetMapping(path = "/range/{id}")
	@Nonnull
	public Mono<Map<HexId, MultiLangTranslation>> getRange(
			@ApiParam(value = "Sample Id within the provided range to fetch", required = true) @PathVariable(name = "id") @Nonnull HexId theId,
			@ApiParam(value = "Type of Xentis IDs Range to fetch translations for", required = true) @RequestParam(name = "range") @Nonnull RangeType theRange,
			@ApiParam(value = "Types of translations to get", required = true, allowMultiple = true) @RequestParam(name = "type", required = true) @Nonnull Set<TranslationType> theTypes) {
		return translationService.getRange(theId, theRange, theTypes);
	}
}
```

Docklet:

```
@Configuration
@EnableSwagger2WebFlux
public class SwaggerConfig {

	@Autowired
	private TypeResolver resolver;

	@Bean
	public Docket api() {
		return new Docket(DocumentationType.SWAGGER_2) //
				.ignoredParameterTypes(ServerWebExchange.class)
				.select()
				.apis(RequestHandlerSelectors.any())
				.paths(PathSelectors.any())
				.build()
				.apiInfo(apiEndPointsInfo())
				.groupName("Basic Data")
				.directModelSubstitute(HexId.class, HexIdModel.class)
				.alternateTypeRules(
						new RecursiveAlternateTypeRule(
								resolver,
								Arrays.asList(
										AlternateTypeRules.newRule(
												resolver.resolve(Mono.class, WildcardType.class),
												resolver.resolve(WildcardType.class)),
										AlternateTypeRules.newRule(
												resolver.resolve(ResponseEntity.class, WildcardType.class),
												resolver.resolve(WildcardType.class)))))
				.alternateTypeRules(
						new RecursiveAlternateTypeRule(
								resolver,
								Arrays.asList(
										AlternateTypeRules.newRule(
												resolver.resolve(Flux.class, WildcardType.class),
												resolver.resolve(List.class, WildcardType.class)),
										AlternateTypeRules.newRule(
												resolver.resolve(ResponseEntity.class, WildcardType.class),
												resolver.resolve(WildcardType.class)))));
	}

	private ApiInfo apiEndPointsInfo() {
		return new ApiInfoBuilder().title("Basic Data API")
				.description("APIs to handle Basic Data (i.e. Instrumens, Funds, Portfolios, Prices, Key Figures, etc.)")
				.contact(new Contact("Xentis Java Team", "", "XENTISJavaDev@profidatagroup.com"))
				.license("XENTIS")
				.version("0.0.1")
				.build();

	}

	//For Swagger Editor compatibility.Check:https://github.com/swagger-api/swagger-core/issues/2944
	@Bean
	public JacksonModuleRegistrar swaggerJacksonModuleRegistrar() {
		return objectMapper -> ReferenceSerializationConfigurer.serializeAsComputedRef(objectMapper);
	}
```

the ObjectA, aka the MultiLangTranslation:

```
public class MultiLangTranslation extends HashMap<String, Translation> {

	/**
	 * The long <code>serialVersionUID</code>.
	 */
	private static final long serialVersionUID = 1360228790913298738L;
	//Marker class
}
```

The Translation:

```
public class Translation implements Serializable {

	/**
	 * The long <code>serialVersionUID</code>.
	 */
	private static final long serialVersionUID = -26163810284100442L;

	private String small;
	private String medium;
	private String big;
        .......
```

and the HexId:

```
public class HexId implements Serializable {

	private static final long serialVersionUID = 201711281450L;

	private long id;

	public HexId(String theHexValue) {
		this(Long.parseLong(theHexValue, 16));
	}

	public HexId(long theId) {
		id = theId;
	}

	public String getValue() {
		return Long.toHexString(id).toUpperCase();
	}

	public Long getId() {
		return id;
	}
```

Since my Service had 3 methods I was getting the Translation and the MultiLangTranslation repeated 3 times.
During my debug I was finding out that at a method (not finding it now in the old snapshot) which tries to find the root models from the Mono<Map<HexId, MultiLangTranslation>> it was only returning the HexId object, since the MultiLangTranslation extended from Map.

But! This also happened for other services where the principal objects didn't extended from Map, but instead referenced directly or indirectly this MultiLangTranslation.

Also for another service where the returns were only: Mono<Map<HexId, List<HexId>>> is only generated one HexId model, but it didn't allow to make the directModelSubstitute(HexId.class, HexIdModel.class), as it would throw an exception saying that HexId would have been already registered.

Sory for the huge information, hope it can help..

MaksimOrlov:
@joao-rebelo thanks a lot for this infromation, Could you please add the generated json schema file. And could you, please, provide full code listing for  **Translation** class with getters and setter.  As I can see class *HexId* is not symmetric (difference between getters and setters). So for serialization and deserialization will be generated **two** models.

joao-rebelo:
Follows the translation. I believe it is symmetric.

```
public class Translation implements Serializable {

	/**
	 * The long <code>serialVersionUID</code>.
	 */
	private static final long serialVersionUID = -26163810284100442L;

	private String small;
	private String medium;
	private String big;

	public Translation() {
		//FOR SPRING DESERIALIZATION ONLY
	}

	public Translation(String theMedium) {
		small = null;
		medium = theMedium;
		big = null;
	}

	public Translation(String theSmall, String theMedium, String theBig) {
		small = theSmall;
		medium = theMedium;
		big = theBig;
	}

	public String getSmall() {
		return small;
	}

	public String getMedium() {
		return medium;
	}

	public String getBig() {
		return big;
	}

	@Override
	public String toString() {
		return new StringBuilder("{SMALL=").append(small).append(", MEDIUM=").append(medium).append(", BIG=").append(big).append("}").toString();
	}
}
```

The generated doc:

[api-docs.txt](https://github.com/springfox/springfox/files/3237189/api-docs.txt)

MaksimOrlov:
Ok, I will make test based on your listing and then fix everything, that looks like ambiguous at the swagger output file. Will back to you in a week.

PR #2069, opened 2017-10-16
Optimise HandlerMethodResolver.getMemberMethods

simongajdosech:
Hello, this is a small optimization, but it greatly reduced startup time of our server. Please review. Thanks.

#### What's this PR do/fix?
Time of generating documentation greatly increases with number
of methods in controller (polynomial growth I think). This fixes the problem by caching
the resolved methods for already processed controller classes.

No, this doesn't change functionality, only caches results to a single method call. Tested this manually.

In our project we have controllers with lots of methods. Generating documentation for these was very slow. This greatly improved startup time of our server - 100 seconds down to 40 seconds.


dilipkrish:
Thank you for this PR! 🤘

PR #2081, opened 2017-10-28
Fix child expansion context creation

gzsombor:
Because currently it modifies the parent expansion context's seenTypes collections,

this cause problems with parameters like:
```
 class SomeFilter {
   LongFilter id;
   LongFilter user;
 }
```

In this case, the user field was not expanded.

 This issue is found in  #https://github.com/jhipster/generator-jhipster/issues/6543


dilipkrish:
This is awesome. Who'd have thought of this scenario :) Thanks for fixing it 🤘. How about about also adding a couple tests?

gzsombor:
I've added a test case for this, hope it's ok, welcome !

dilipkrish:
Thank you @gzsombor!

Issue #1082, opened 2015-12-03
Replace Guava by java8 predicates function


ouaibsky:
Hi
last version of springfox rely on guava, maily  to apply predicates for path filtering.
It's an issue for a documentation tools like this to depend of a so hudge library. I means we need this dependencies for swagger support for spring mvc endpoint.

This type of features are now supported natively by java 8

Thx
Christophe

dilipkrish:
Currently we support java 6 through 8. So unless we deprecate support for java 6 and 7 thats not possible at least in the near future.

ouaibsky:
Hi
Thx, it's a shame because guava was not part of swagger 1

Is it possible to envisage something like this:
* Design filtering interface
* Mark guava as an optionnal dep in pom
* Provide an implementation of filtering that check if java 8 is present, if not fall-back on guava

Christophe

dilipkrish:
Absolutely possible! Once we move to java 8 there won't be a need to use explicit guava/java 8 types even. They could just be replaced by lamdas.

Unfortunately, like I said before, we still support java 6 and spring 3.2 and above. So thats not going to happen anytime soon. 

Also needless to say the cost of doing that might be quite high in terms of time. I do this in my spare time and no one is paying me to maintain this library. I don't have cycles to work on large refactoring efforts when nothing is really broken.

ouaibsky:
We all have a limited capex when it's private project ;)

I mean If we propose a PR in this direction is there a chance that it will be accepted ?

dilipkrish:
Abolutely!

RobWin:
You can still use Springfox just as a testCompile dependency and create the Swagger JSON in a unit test and include it into your jar then serve it as static content.
See https://github.com/Swagger2Markup/spring-swagger2markup-demo
No need to include the library into your production code.

cleankod:
@ouaibsky where are you on this PR?

dilipkrish:
@cleankod There have been no takers for it

neil1hart:
I can attempt this.  Some of this is straight forward, cache loading is a little harder, ImmutableSet will be lost...  The big questions is what should be done with VisibleForTesting annotation?

dilipkrish:
@neil1hart `VisibleForTesting` is just a marker annotation to signal why a particular method is not private/restricted than needed. It can go away. Also keep in mind there is also #2086 that you can build on.

neil1hart:
[...]

PR #2082, opened 2017-10-29
Remove guava and use java 8 constructs (#1082)

copa2:
This PR removes the guava library dependency with java 8 constructs.
(Fix issue #1082)

The main reason is that other code also uses guava in a incompatible version and guava is not always binary compatible. 

Incompatiblilites: 
* This will update jdk from 1.6 to 1.8 as minimum version
* Binary incompatiblities 
* Public API changes where guava constructs have been exposed

What is missing:
* Documentation updates (will follow / references to guava)
* Cleanups (Predicates, Function can be written more compact with java 8)

Checked alternatives:
* Using shading plugin to shade guava. This makes each sub-project artifact big due whole guava library has to be shaded into the artifact.


dilipkrish:
Thank you @copa2. Will definitely pull in this change after the 2.8.0 release. Currently the library supports down to java 6 and this is a breaking change => up rev to 3.0.0 after this release.

copa2:
Hi dilipkrish

Just checked the failure I receive from circleci. 
Problem is that the CachingOperationReader already today is probably not doing what is expected.
https://github.com/springfox/springfox/blob/ec9af577593c3565bf57a00c13e4e89d7cb804ef/springfox-spring-web/src/main/java/springfox/documentation/spring/web/scanners/CachingOperationReader.java#L57

Each entry receives a new OperationCachingEquivalence which calls the Wrapper. 
Within the Wrapper the equals checks if the Equivalence is the same Object
`if (this.equivalence.equals(that.equivalence)) {` before it call the doEquivalent.

copa2:
This PR is replaced by PR #2086, due unneeded crlf.

PR #2101, opened 2017-11-14
Fix Remapping issue

apixandru:
The PropertySourcedMapping annotation was not being used as intended (I think)


dilipkrish:
@apixandru thanks for catching that! 👍

PR #2103, opened 2017-11-15
Fix some tests under windows

apixandru:
tests were failing under windows because of line feed characters


dilipkrish:
@apixandru thank you!

PR #2144, opened 2017-12-12
Allow ApiModelProperties on methods to be discovered from superclasses

RoyJacobs:
#### What's this PR do/fix?
Methods that have ApiModelProperty annotations on superclasses (or interfaces) were not discovered in all cases because `ApiModelProperties#findApiModePropertyAnnotation` only looks at direct annotations.

Yes

`ApiModelPropertyPropertyBuilder` seems to correctly discover annotations when the annotated element is missing, but this is not the case otherwise.

I did not create an issue.

### DIFF
--- a/springfox-swagger-common/src/main/java/springfox/documentation/swagger/schema/ApiModelProperties.java
+++ b/springfox-swagger-common/src/main/java/springfox/documentation/swagger/schema/ApiModelProperties.java
@@ -34,6 +34,7 @@ import springfox.documentation.service.AllowableValues;
 import springfox.documentation.spring.web.DescriptionResolver;
 
 import java.lang.reflect.AnnotatedElement;
+import java.lang.reflect.Method;
 import java.util.Collections;
 import java.util.List;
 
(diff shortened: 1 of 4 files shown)


dilipkrish:
@RoyJacobs thank you!

RoyJacobs:
@dilipkrish Thanks for merging! Unfortunately I've run into another case where it's not working correctly: In case the annotated element in `ApiModelProperties#findApiModePropertyAnnotation` is a field then my fix will still not work (because then the annotated element is of type `Field`, not of type `Method`). Is there a way to configure Springfox to ignore fields and only look at getters?

dilipkrish:
There cannot be fields on interfaces. The field behavior should just work I believe.

RoyJacobs:
Sorry, I mean the opposite: when there is a field and a getter inside the
implementing class. In that situation the annotation on the field is used
(and missing) whereas I would like the annotation on the getter to be used,
regardless whether the getter is on the interface or on the implementing
class.

Op do 14 dec. 2017 13:04 schreef Dilip Krishnan <notifications@github.com>:

dilipkrish:
You can alter that behavior by configuring jackson.

RoyJacobs:
I did not realize this is related to Jackson configuration. I will investigate, thanks!

dilipkrish:
Using [this technique](https://stackoverflow.com/a/7108530/19219) or using using an annotation.

PR #2160, opened 2017-12-20
Fix conversion of byte to integer with max / min

avdv:
#### What's this PR do/fix?

A single Byte/byte is represented as a (base64 encoded) String in the generated swagger API (type: `string`, format: `byte`).

This mapping does not work.

When using a Client generated with swagger-codegen, this fails miserably because for a single value, the generated client expects an Array.

Here is a sample entity class:
```kotlin
@Entity data class Sample(val prop: Byte = 1)
```
For this class the server sends out a JSON object as follows:
```json
{
  "prop": 1
}
```
Alas, the api-spec generated by springfox-swagger is:
```yaml
prop:
     type: string
     format: byte
```


dilipkrish:
Very dramatic issue description, I must say "fails miserable", "alas" 🥇 I'm surprised no one has caught it for so long. I'll pull it in shortly. Thanks for the PR 🤘!

dilipkrish:
Thank you @avdv!

PR #2189, opened 2018-01-16
Should use credentials on same origin when fetching springfox resources

damiengenet:
If swagger-ressources is not publicly accessible Swagger UI loading will fail, even though the current user is already authenticated.

This is caused during the SpringFox initialization by missing request credentials (for instance a session cookie) while trying to load `/swagger-resources/configuration/ui`, `/swagger-resources/configuration/security` and `/swagger-resources`. Swagger UI itself does properly reuse existing credentials afterwards.

This is a regression in 2.8.0 as this worked fine with springfox-swagger-ui 2.7.0.

### DIFF
--- a/springfox-swagger-ui/src/web/js/springfox.js
+++ b/springfox-swagger-ui/src/web/js/springfox.js
@@ -2,13 +2,13 @@ window.onload = () => {
 
   const buildSystemAsync = async (baseUrl) => {
     try {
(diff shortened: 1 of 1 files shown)


Hronom:
What the status of this?

dilipkrish:
@damiengenet Thanks for the PR!

Issue #2257, opened 2018-02-20
Tests fail on windows


atsu85:
- [x] What version of the library are you using?
latest master (commit b1cf952), but the same applies for version 2.8.0 tag (commit b37d9c0)

What kind of issue is this?

 - [x] Bug report.
    - there are test failures when running tests on Windows (using `gradlew clean build`). Doesn't matter if using Java 8 or java 7 (also tried Java 6, but that didn't work because of Gradle version)

# Failing tests:
- [ ]    FunctionContractSpec. should honor api v1.2 contract [declaration-concrete-controller.json] at endpoint [/default/concrete-controller] (fixed by: [e83be4c](https://github.com/atsu85/springfox/commit/e83be4c8b2394e9dc491e96fe1b2a264ed6e9e00))
- [x]    FunctionContractSpec. should honor api v1.2 contract [declaration-feature-demonstration-service.json] at endpoint [/default/feature-demonstration-service]
- [ ]    FunctionContractSpec. should honor api v1.2 contract [declaration-pet-service.json] at endpoint [/default/pet-service] (fixed by: [e83be4c](https://github.com/atsu85/springfox/commit/e83be4c8b2394e9dc491e96fe1b2a264ed6e9e00))
- [x]    FunctionContractSpec. should honor swagger resource listing
- [x]    FunctionContractSpec. should honor swagger v2 resource listing bugs
- [x]    FunctionContractSpec. should honor swagger v2 resource listing bugsDifferent
- [x]    FunctionContractSpec. should honor swagger v2 resource listing concrete
- [x]    FunctionContractSpec. should honor swagger v2 resource listing featureService

atsu85:
I also uploaded [zip file of swagger-contract-tests/build/reports/tests folder](https://github.com/springfox/springfox/files/1741745/build_reports_tests_test-b1cf952-win-java7.zip) (built on Windows 10 with java 7) - You could inspect the failure details after unzipping and opening index.html

dilipkrish:
Are you using the gradle wrapper? => `/.gradlew clean build`

atsu85:
hi, yes:
> `gradlew clean build`

later i also tried the same command that is in the circle.yml and that also produced the same test failures

PR #2258, opened 2018-02-20
Issue 2257 fix default file encoding related failures

atsu85:
#### What's this PR do/fix?
This PR fixes two tests that fail on windows (reported with #2257)
No production code, just fix for unit test
This only fixes some tests reported with #2257, there are other tests that still fail on windows (doesn't make any other previously successful tests fail)
#2257


dilipkrish:
Thank you @atsu85!!

Issue #2276, opened 2018-03-02
@RequestParam with defaultValue of empty string on String parameter marked as required


matthewmcgarvey:
Came across a query parameter showing that it is required when, in the spring controller, it has a defaultValue of an empty string.  Looking through the repository, I see that defaultValue sans required false is supported but the code is checking to see if the string is not blank in addition to checking if the value is not the `DEFAULT_NONE` that is the actual default value on that field.

My contention is that any value other than `DEFAULT_NONE` should be considered valid.

Example that creates the bug:
```java
@GetMapping
public String search(@RequestParam(defaultValue = "") String email) {
    // The Request Param is implicitly not required because of the defaultValue (which is handled
    // by springfox), but the defaultValue is an empty string to avoid null checks
    System.out.println("Email should be an empty string if not provided: " + email);
    return email;
}
```
File To Change: https://github.com/springfox/springfox/blob/master/springfox-spring-web/src/main/java/springfox/documentation/spring/web/readers/parameter/ParameterRequiredReader.java

PR #2277, opened 2018-03-02
Not checking if defaultValue on is empty

matthewmcgarvey:
Fixes issue #2276


dilipkrish:
@matthewmcgarvey thanks for the PR! Sorry I did not notice it in time. I implemented the change that you requested however!

Issue #2311, opened 2018-03-19
springfox-data-rest: Inconsistent model returned between findAll and search methods


iles-2e:
When using the latest snapshot (2.8.1-SNAPSHOT) of springfox-data-rest, I found an inconsistent behavior between _findAll_ and _search_ methods. Both methods return _List\<Entity\>_ and produce the same JSON HAL similar to this:

`{
  "_embedded" : {
    "someResourceRel" : [ {entities}]
  },
  "_links" : {...`

but springfox reports different models used. 

For _findAll_ method a _Resources\<<Entity\>>_ is reported while for _search_ method a _Resources\<<List\<<Entity\>>\>>_ is reported. 

In the contract for **declaration-spring-data-rest.json** it is also defined a similar example with _ResourcesOfPerson_ and _ResourcesOfListOfPerson_ while it should probably produce the same result.

[Resources](https://docs.spring.io/spring-hateoas/docs/current/api/org/springframework/hateoas/Resource.html) is already a collection wrapper... I doubt it wraps a collection of additional collections.

I would gladly contribute a PR if this proves to be a bug indeed. Any thoughts?

dilipkrish:
Seems like a duplicate of #1980

iles-2e:
Not quite... #1879,  #1927 and #1980 report the different output format, i.e. the serialization of [Resources](https://docs.spring.io/spring-hateoas/docs/current/api/org/springframework/hateoas/Resource.html) which can almost be corrected using the Spring HATEOS object mapper (_halObjectMapper bean with [Jackson2HalModule](https://docs.spring.io/spring-hateoas/docs/current/api/org/springframework/hateoas/hal/Jackson2HalModule.html)) but that is a different subject. 

What this bug report is trying to address is the wrong content type inside the [Resources](https://docs.spring.io/spring-hateoas/docs/current/api/org/springframework/hateoas/Resource.html) instance.

The [EntityFindAllExtractor](https://github.com/springfox/springfox/blob/master/springfox-data-rest/src/main/java/springfox/documentation/spring/data/rest/EntityFindAllExtractor.java) @ line 71 resolves the return type using the [RepositoryMetadata.getReturnedDomainClass](https://docs.spring.io/autorepo/docs/spring-data-commons/1.9.2.RELEASE/api/org/springframework/data/repository/core/RepositoryMetadata.html) method which extracts the domain type from the given collection.
[...]

PR #2313, opened 2018-03-20
springfox-data-rest: fix EntitySearchExtractor return type

iles-2e:
#### What's this PR do/fix?

This is a fix for #2311 

_declaration-spring-data-rest_ contract has been updated to reflect the changed behavior.

_findAll_ and _search_ methods should return the same type of result.

#2311


dilipkrish:
Thank you @iles-2e

PR #2323, opened 2018-03-25
fix for spring data rest 3.0.5.RELEASE while maintaining backward com…

deodeveloper:
…partibility

#### What's this PR do/fix?
The current swagger version doesn't work with new version of spring data rest (3.0.5.RELEASE). This pull request fixes that problem while maintaining backward compatibility.
No new unit tests required for the changes, the existing unit tests have those.
Swagger data rest doesn't work with new spring data rest version 3.0.5.RELEASE due to the fact that new spring version returns java 8 Optional<T> instead of direct object.
Current Swagger data rest (2.8.0) doesn't work with new spring data rest version 3.0.5.RELEASE


dilipkrish:
Thanks @deodeveloper! 🤘

Issue #1575, opened 2016-11-21
Parameter Annotations do not traverse super classes for resolution


dilipkrish:

yummish:
I think I have a similar issue:

### with the interface generated by swagger I have:
`@Api(value = "lawyers", description = "the lawyers API")
public interface LawyersApi {

    @ApiResponses(value = { 
        @ApiResponse(code = 200, message = "kkj", response = StatsData.class) })
    @RequestMapping(value = "/lawyers/{lawyerId}/stats/case_compensation",
        produces = { "application/json" }, 
        method = RequestMethod.GET)
    StatsData getCaseCompensationStatsByLawyerId(
@ApiParam(value = "",required=true ) @PathVariable("lawyerId") String lawyerId
);
}
`
### and controller:
@RestController
public class LawyersController implements LawyersApi {

  @Override
  public StatsData getCaseCompensationStatsByLawyerId( String lawyerId) {
    // TODO Auto-generated method stub
    System.out.println(lawyerId);
    return null;
  }
}

If i call this api, then the lawyerId is identified as query parameter, not the path variable as defined in the interface. To resolve this, I have to copy interface's parameters' annotations to the implementation.

So, I'm just wondering if I'm missing anything or is this relates to this issue.

thanks

Morl99:
@dilipkrish: I would like to +1 this, giving an explanation alongside. 
Our ideal flow of using swagger is specification first. We currently specify an api.yaml document, generate interfaces, implement controllers (or change the existing controllers along the changes in the interfaces), and the running service then exposes the current specification through springfox. 
We have a lot of parameter description, explaining how the parameters work, but this information cannot be exposed to service consumer at the moment.

If I can do something to help solving this issue let me know, but I think that I will probably not be able to fix this with a PR, as I am not familiar with the implementation details of the springfox project.

swatisacheti:
Any timeline on when this will be fixed? Having the same issue as [388](https://github.com/springfox/springfox/issues/388). The API doc still does not display descriptions of method parameters defined by @ApiParam in the interface. We're using version 2.7.0

dilipkrish:
@Morl99 would definitely appreciate help with fixing this. The only caveat to this issue is that you can run into a trap of traversing multiple interfaces as a result. If you can perhaps do that based on a configuration in the docket that would be awesome.

Mulder8472:
> The only caveat to this issue is that you can run into a trap of traversing multiple interfaces as a result

[...]

PR #2331, opened 2018-03-28
Fix for Parameter Annotations like @ApiParam do not traverse super cl…

flowerrrr:
…asses for resolution #1575.

#### What's this PR do/fix?
Annotations of method parameters are now resolved by traversing super classes and interfaces.
Annotations are cached per handler method to avoid repeated lookups.
The code to find the inherited annotations is taken from swagger-core's ReflectionUtils class.

The first commit of this PR should be considered a draft. Feedback welcome.

No unit tests yet. 
I'm happy to provide some tests but need some help doing so.

### DIFF
--- a/springfox-spring-web/src/main/java/springfox/documentation/spring/web/readers/operation/HandlerMethodResolver.java
+++ b/springfox-spring-web/src/main/java/springfox/documentation/spring/web/readers/operation/HandlerMethodResolver.java
@@ -38,6 +38,7 @@ import org.springframework.core.MethodParameter;
 import org.springframework.core.ParameterNameDiscoverer;
 import org.springframework.web.method.HandlerMethod;
 import springfox.documentation.service.ResolvedMethodParameter;
+import springfox.documentation.spring.web.readers.operation.util.AnnotationUtils;
 
 import java.lang.reflect.Method;
 import java.lang.reflect.ParameterizedType;
@@ -178,9 +179,10 @@ public class HandlerMethodResolver {
         MethodParameter[] methodParameters = methodToResolve.getMethodParameters();
         for (int i = 0; i < input.getArgumentCount(); i++) {
           parameters.add(new ResolvedMethodParameter(
-              discoveredName(methodParameters[i]).or(String.format("param%s", i)),
-              methodParameters[i],
-              input.getArgumentType(i)));
+                  methodParameters[i].getParameterIndex(),
+                  discoveredName(methodParameters[i]).or(String.format("param%s", i)),
+                  AnnotationUtils.getInheritedAnnotations(methodToResolve, methodParameters[i].getParameterIndex()),
+                  input.getArgumentType(i)));
         }
         return parameters;
       }
--- a/springfox-spring-web/src/main/java/springfox/documentation/spring/web/readers/operation/util/AnnotationUtils.java
+++ b/springfox-spring-web/src/main/java/springfox/documentation/spring/web/readers/operation/util/AnnotationUtils.java
@@ -1,0 +1,49 @@
+/*
+ *
+ *  Copyright 2016 the original author or authors.
+ *
+ *  Licensed under the Apache License, Version 2.0 (the "License");
+ *  you may not use this file except in compliance with the License.
+ *  You may obtain a copy of the License at
+ *
+ *         http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *  Unless required by applicable law or agreed to in writing, software
+ *  distributed under the License is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ *
+ *
+ */
+
+package springfox.documentation.spring.web.readers.operation.util;
+
+import org.springframework.util.ConcurrentReferenceHashMap;
+import org.springframework.web.method.HandlerMethod;
+
+import java.lang.annotation.Annotation;
+import java.util.Arrays;
+import java.util.List;
+import java.util.Map;
+
+public class AnnotationUtils {
+
+    private static final Map<HandlerMethod, Annotation[][]> ANNOTATION_CACHE =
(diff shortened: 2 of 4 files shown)


dilipkrish:
@cramja there are no tests however

cramja:
@dilipkrish I could cherry-pick the commit and add coverage if needed. Waiting for @flowerrrr 's feedback. In the mean time I found out that spring does not map `@RequestParam` annotations from interface methods to the overriding methods, therefore we're going with a different approach. 

Still, this PR seems useful if completed.

dilipkrish:
@cramja that was one of the reasons Im skeptical about this PR. Also it is potentially a performance bottleneck to traverse inheritance chains to determine annotations. That is why it was left as an exercise for the library consumer to implement as a plugin.

flowerrrr:
I'm sorry, I've no resources to continue work on this issue. 



    Am Montag, 25. Juni 2018, 16:25:14 MESZ hat Marc S <notifications@github.com> Folgendes geschrieben:  
 
 
Hello, I'm working in a project that separates the REST interfaces from the controller classes that implement them but a nagging issue is that the annotations on the interface do not work. I tested this PR against our code and it fixes the issue. @flowerrrr any plans to continue work on this PR? I'd be happy to contribute tests/fixes if needed.

cramja:
No worries @flowerrrr . I will look into adding this functionality as a plugin if we end up needing it. In the mean time, we found a work around:

```
interface FooInterface {

  @ApiOperation("Gets Foo")
  @GetMapping("v1/foo/{id}")
  default Foo getFoo(@RequestParam @ApiParam String id) {
    return getFooImpl(id);
  }

  Foo getFooImpl(String id);
}
```

This pattern recovers all the annotations without code duplication.

dilipkrish:
Thats a neat trick! thanks @cramja. However I can imagine this will quickly get unwieldy for a sufficient number of operations

cramja:
@dilipkrish I expect it will be cumbersome as we add more methods. I see that there is a [SPI package](https://github.com/springfox/springfox/tree/master/springfox-spi/src/main/java/springfox/documentation/spi/service) that we can probably use to do the recursive annotation lookup - double checking- is this what you meant when you said that this could be accomplished by adding a plugin?

dilipkrish:
@cramja yes. You'd have to do what @flowerrrr did travel up the inheritance hierarchy. One of the reasons that is not exposed directly as a plugin extensibility feature is because it relies on the controller being a subclass of an interface that you're searching for, and that is an abstraction thats not always true. For e.g. in spring data rest there are no concrete controllers for each of the domain objects.

flowerrrr:
Could someone please describe in a few words, how this functionality can be implemented? I've read the documentation about Plugins but i'm not sure which of the available Plugin-Interfaces is my starting point.

cramja:
I've read some of the code and I found that the [bootstrapper](https://github.com/springfox/springfox/blob/master/springfox-spring-web/src/main/java/springfox/documentation/spring/web/plugins/DocumentationPluginsBootstrapper.java#L148) is a good place to start reading. Other than that, I have not written a plugin so, can't be of too much help.

flowerrrr:
Here's a solution  using the plugin mechanism.

```
import io.swagger.util.ReflectionUtils;
import lombok.SneakyThrows;
import org.springframework.core.annotation.Order;
import org.springframework.stereotype.Service;
import org.springframework.util.ConcurrentReferenceHashMap;
import org.springframework.web.method.HandlerMethod;
import springfox.documentation.RequestHandler;
import springfox.documentation.service.ResolvedMethodParameter;
import springfox.documentation.spi.DocumentationType;
import springfox.documentation.spi.service.ParameterBuilderPlugin;
import springfox.documentation.spi.service.contexts.ParameterContext;
import springfox.documentation.spi.service.contexts.RequestMappingContext;

import java.lang.annotation.Annotation;
import java.lang.reflect.Field;
import java.util.Arrays;
import java.util.List;
import java.util.Map;

import static springfox.documentation.swagger.common.SwaggerPluginSupport.SWAGGER_PLUGIN_ORDER;

@Service
@Order(SWAGGER_PLUGIN_ORDER - 1) // plugin must be called before ApiParamParameterBuilder
public class InheritedApiParamPlugin implements ParameterBuilderPlugin {

    @Override
    public void apply(ParameterContext parameterContext) {
        RequestMappingContext requestContext = getField(parameterContext.getOperationContext(), "requestContext");
        RequestHandler handler = getField(requestContext, "handler");
        ResolvedMethodParameter resolvedMethodParameter = parameterContext.resolvedMethodParameter();
        List<Annotation> inheritedAnnotations = AnnotationUtils.getInheritedAnnotations(handler.getHandlerMethod(), resolvedMethodParameter.getParameterIndex());
        inheritedAnnotations.stream()
                .filter(annotation -> !resolvedMethodParameter.getAnnotations().contains(annotation))
                .forEach(annotation -> resolvedMethodParameter.getAnnotations().add(annotation));
    }

    @SneakyThrows({NoSuchFieldException.class, IllegalAccessException.class})
    private static <T> T getField(Object target, String fieldName) {
        Field field = target.getClass().getDeclaredField(fieldName);
        field.setAccessible(true);
        return (T) field.get(target);
    }

    @Override
    public boolean supports(DocumentationType delimiter) {
        return true;
    }

    public static class AnnotationUtils {

        private static final Map<HandlerMethod, Annotation[][]> ANNOTATION_CACHE =
                new ConcurrentReferenceHashMap<>(256);

        public static List<Annotation> getInheritedAnnotations(HandlerMethod methodToResolve, int parameterIndex) {
            Annotation[][] parameterAnnotations = getInheritedParameterAnnotations(methodToResolve);
            return Arrays.asList(parameterAnnotations[parameterIndex]);
        }

        private static Annotation[][] getInheritedParameterAnnotations(HandlerMethod methodToResolve) {
            Annotation[][] annotations = ANNOTATION_CACHE.get(methodToResolve);
            if (annotations == null) {
                annotations = ReflectionUtils.getParameterAnnotations(methodToResolve.getMethod());
                ANNOTATION_CACHE.put(methodToResolve, annotations);
            }
            return annotations;
        }

    }

}

```

The code uses a utility function from swagger-core to find annotations in super classes and interfaces.

dilipkrish:
Closing this PR in lieu of #3064

Issue #918, opened 2015-08-16
Allow support for describing alternate types with rich information


dilipkrish:
Currently alternate types only support type substitution. Allow the meta data from the original property/model to flow through. Also provide a way to describe these alternate types in a rich way via docket configuration.

Something like this perhaps
```java

docket.alternateTypeRules(
    richAlternateTypeRule(LocalDateTime.class, String.class)
        .withAdditionalInfo(modelPropertyBuilder().withNotes("blah"))
```

Related to #914

cbornet:
:+1:

rdelaoc:
This is fantastic. It would be great being able to globally set the example for properties of a certain type (e.g. custom date formatting).

dangoslen:
I see this is a pretty old request, but I'd like to take a stab at it if thats ok! Our team would really like the ability to have a global examples, notes, etc for the same reasons listed by others here. 

However, I was wondering about one of the statements at the beginning of this issue. @dilipkrish: could give elaborate on your statement of "Allow the meta data from the original property/model to flow through". Are you saying if you were substituting `ClassA` for `ClassB`, you would want any `ModelProperty` on an instance of `ClassA` to be pushed onto the substituted `ClassB`?

Thanks!

mickster04:
in my scenario, I have datetime objects, I want to be able to have some annotation like (@ApiModelProperty(temporalFormat = "yyyy-MM-dd'T'HH:mm:ss.SSS'Z'")) 
or even, if I could pass in an appropriate simpleDateFormat-esque object (as is correct for whatever object I am working with, whether it's LocaleDateTime, joda.DateTime, or calendar) that will allow me to format the required date object to the format I need it.
Ideally swagger would read the
	@JsonFormat(shape= JsonFormat.Shape.STRING, pattern="yyyy-MM-dd'T'HH:mm:ss.SSS'Z'")
annotation and use that as the directive for how to send the date time field. the problem is when "trying it out" as it sends the date in the wrong format by default.

dilipkrish:
@dangoslen Sorry just noticed I hadn't responded. By meta data flowing through, I mean if you had the following class 

```java
Class A {
    @ApiModelProperty(name="date", notes="this date is of this format ", dataType = "date-time", example = "2015-01-31T08:00:00")
    LocalDate date
}
```

And if you had 

```java
    docket.alternateTypeRules(
          richAlternateTypeRule(LocalDate.class, String.class))
```

[...]

PR #2351, opened 2018-04-10
Rich alternate type rule

dangoslen:
#### What's this PR do/fix?
This PR attempts to address the first part of issue #918, which creates the ability to retrieve properties as if it were the original type instead of the alternate.

I believe I have added the tests needed.

I tried to keep changes small at first to incorporate feedback sooner rather than later. I'm open to improvements for sure!

#918


dilipkrish:
@dangoslen Just getting around to reviewing it. 

I was wondering, since propertyHost is a new method (additive not breaking), why not have AlternateTypeRule remain as it is (i.e. not turn it into an interface) and just add the new method to it.

And keep everything else the same, like you have it. The reason Im suggesting that is 
1. it maintains backwards compatibility and i could pull it in before 2.9 which I plan to release this weekend or early next week
2. Still introduce the new behavior of exposing the original object.
3. Also, I like the idea of introducing the interface, but Im a little ambivalent about making Simple/Rich alternate type rules public. I'd rather just expose a factory method that creates either one of the interface.

I could be swayed either way, let me know what you think? Also would you mind squashing your merge commit?

dangoslen:
Fair points. No need for another layer of abstraction there. Now that I re-look, I think I got halfway through and thought the same thing, and forgot to back it out. 

Per point 3, I _thought_ the `AlternateTypeRule` was public already... was it not. I do like the idea of static factory methods though, which reminds me I never added the helper function to create a `RichAlternateType` on the `Docket`. I'll at least get the helper methods added to the `Docket` as that was included in the original feature request anyhow.

Also, and this is where my newb status to OpenSource comes in, how would I squash the merge commit after its been committed? I've tried to follow a few articles using `git rebase -i`, but to no avail. Is the best option to reset to before the merge and run a new merge command? Sorry for the newb status

dilipkrish:
@dangoslen 

> Sorry for the newb status

We're all newbs sometime or the other. I still learn things about git and github every day. Sorry I assumed you'd know. The reason is the quality of your PR; each commit tells a story. It makes reviewing so easy. It is also why the empty merge commit stands out :)

I believe if you do the following it should work. That should remove the empty commit.

```bash
# Assumes your current branch is the one you're working this PR on 
# and you dont have any local changes that are not committed.
git fetch origin master
git rebase origin/master
```

Yeah  `AlternateTypeRule` was created when I didn't put much thought into whether something was public or not :) In OSS you learn pretty fast that thats not a great idea. There is a [factory/helper however](https://github.com/springfox/springfox/blob/5ce78b962a0a95edddcb29e676b80cb7163bf87b/springfox-core/src/main/java/springfox/documentation/schema/AlternateTypeRules.java#L37-L59) that tries to make it easier to create these rules with the right order.

dangoslen:
👍 Thanks for the help there. Wasn't aware git could do a merge and then a rebase after the fact. 

Ironically, after getting to the point where I wanted to test out the `Docket` class and do final validation that using the `RichAlternateTypeRule` returned the right swagger .json, it looks like the feature exists using a normal `AlternateTypeRule` (or direct substitution). 

I added a new field of `LocalDate` in the `User` model in the petstore example and then added the `@ApiModelProperty` to the field and it had the right values without modifying the `Docket` in `Swagger2SpringBoot` example configuration. Maybe I misunderstood the feature request.

With that being said, maybe this PR is the wrong way to go and I should switch to try to get the defaults for a property added (which is the second piece mentioned in the original request). 

Thoughts?

dilipkrish:
> I added a new field of LocalDate in the User model in the petstore example and then added the @ApiModelProperty to the field and it had the right values without modifying the Docket in Swagger2SpringBoot example configuration. 

Im surprised that it does. Perhaps it was fixed since the issue was created. The initial idea was to supplement additional metadata using fluent builders as opposed to referring annotations, which involves few reflection calls. I'll revisit the code and see what is missing.

PR #2404, opened 2018-05-08
Using media type to determine correct param type when expanding param…

### LINKED ISSUE #2376: Model attribute expansion should respect form/query parameter type
Reported https://github.com/springfox/springfox/issues/680#issuecomment-336436807 by @smthpickboy

The expanded parameters are always coming through with parameter type of query. It should inspect the consumes media type and determine the type.

andyRokit:
**What's this PR do/fix?**
This PR attempts to resolve #2376 by using the media type to determine the parameter type when expanding.

**Are there unit tests? If not how should this be manually tested?**
The intentional change was made to `OperationParameterReaderSpec.groovy`.  I split the "Should expand ModelAttribute request params" test into two.  One for multipart, one for non-multipart.  I aimed to keep any other test changes to a minimum.

**Any background context you want to provide?**
N/a

**What are the relevant issues?**
#2376


dilipkrish:
Thanks @andyRokit for the PR! 🤘

Issue #2428, opened 2018-05-18
description = "@Size: Min - Max (until #1244 gets fixed)" when param annotated with @Min and/or @Max but #1244 is already fixed


haelduksf:
2.9.0 still contains the above message but #1244 was fixed in 2.5. Will submit a fix shortly.

dilipkrish:
Fixed via #2429

PR #2429, opened 2018-05-18
Remove output related to fixed issue (#1244) #2428

haelduksf:
I noticed some descriptions in my documentation like `@Size: Min - Max (until #1244 gets fixed)`. Looks like that was fixed back in 2.5, and  I see the correct "minimum" and "maximum" values for all of them, so I opened #2428 and got rid of the generated message.


dilipkrish:
Thanks for the PR @haelduksf !!

PR #2455, opened 2018-05-31
Remove the default sort of dockets shown in the home page

tbwork:
This is just a modification with just two slashes, but it means a lot. It fits a design principle that "do not modify something behind users". We known it intends to sort automatically for users. However it will confuse the user due to that the final order of dockets are different from the order specified by user. 
For example, the user defines three dockets with following order:
```
@Bean Docket1().....
@Bean Docket2().....
@Bean Docket3().....
```
But maybe the order shown in the home page become something below:
```
Docket3
Docket1
Docket2
```
This is confusing.
The best solution is removing this line of codes. Let the users decide the order by which dockets are shown.

### DIFF
--- a/springfox-swagger-common/src/main/java/springfox/documentation/swagger/web/InMemorySwaggerResourcesProvider.java
+++ b/springfox-swagger-common/src/main/java/springfox/documentation/swagger/web/InMemorySwaggerResourcesProvider.java
@@ -27,8 +27,7 @@ import springfox.documentation.service.Documentation;
 import springfox.documentation.spring.web.DocumentationCache;
 import springfox.documentation.spring.web.plugins.Docket;
 
-import java.util.ArrayList;
-import java.util.Collections;
+import java.util.ArrayList; 
 import java.util.List;
 import java.util.Map;
 
@@ -75,7 +74,7 @@ public class InMemorySwaggerResourcesProvider implements SwaggerResourcesProvide
         resources.add(swaggerResource);
       }
     }
-    Collections.sort(resources);
+    //Collections.sort(resources);
     return resources;
   }
 


dilipkrish:
@tbwork rather than remove it, I'd rather have intelligent defaults and provide an extensibility to override.

tbwork:
@dilipkrish  I just know you would say like that. I do agree with that we should do much more to let users do less. But in this situation, the cost of writing a sort algorithm for docket list is so small that there is no need to provider intelligent defaults. It likes a tricky balance between love and free. Love users more? or free them and make the logic be easy to understand.

zeko868:
When I applied proposed files changes of @tbwork , the dockets were still displayed sorted alphabetically in the drop-down menu on Swagger UI page. What did the trick was commenting `Stream.sorted(Comparator)` method call on the stream of docket collection inside of `bootstrapDocumentationPlugins` method in `AbstractDocumentationPluginsBootstrapper` class - that was required change on the version where HEAD was pointing to the same commit as current origin/branch (#fb478cf).
On the other hand, when HEAD was on commit #a32a47c, following lines within `start` method inside of `DocumentationPluginsBootstrapper` class has to be replaced:
```
            List<DocumentationPlugin> plugins = pluginOrdering()
                    .sortedCopy(documentationPluginsManager.documentationPlugins());
```
with the following:
```
            List<DocumentationPlugin> plugins = StreamSupport.stream(documentationPluginsManager.documentationPlugins()
                    .spliterator(), false)
                    .collect(toList());
```

tbwork:
It has been a long time, and I guess this is already fixed :)

dilipkrish:
I dont know that it is... plugins continue to be sorted by name. Is it important to make that extensible/custom sortable?

tbwork:
> I dont know that it is... plugins continue to be sorted by name. Is it important to make that extensible/custom sortable?

Hi, dilipkrish. I was too busy to pay attention to this issue. About " Is it important to make that extensible/custom sortable?", I think it is. However it is just a tiny improvement, feel free to make the decision :). Nice Day!

Issue #2446, opened 2018-05-27
ApiParam allowableValues string with spaces is not supported


uriparush:
Hi,
APiParam annotation has a parameter allowableValue.
This takes a comma separated string. But if i have a space in one of the values, it is stripping it off. 
Which is not good. I can pass a string with spaces from the browser url. So i should be able to pass it from swaagger UI as well.

I think the issue is with this line
I found the below line in class "ApiModelProperties" 
String trimmed = allowableValueString.trim().replaceAll(" ", "");  // line 67

This issue was already discovered in previous versions (Issue number #1211) and was fixed for version 2.6.0 and in some future versions the bug was recreated.

I'm currently using version 2.8.0.

Thanks

dilipkrish:
@uriparush since you've researched the issue a bit would you be able to submit a PR?

ctruzzi:
I've created a PR that should resolve this

uriparush:
@dilipkrish sorry for not handling I was on vacation, next time I will create a PR.
Thanks

Issue #1211, opened 2016-03-09
ApiParam allowableValues string with spaces


pvpkiran:
APiParam annotation has a parameter allowableValue. This takes a comma separated string. But if i have a space in one of the values, it is stripping it off. Which is not good. I can pass a string with spaces from the browser url. So i should be able to pass it from swaagger UI as well. 

I think the issue is with this line
` allowableValueString = allowableValueString.trim().replaceAll(" ", "");`
in `ApiModelProperties.java` file.  Line 59

dilipkrish:
Thanks for catching that @pvpkiran :metal:

uriparush:
Hi,
I have the same issue and I use version 2.8.0
I found the below line in class "ApiModelProperties"
String trimmed = allowableValueString.trim().replaceAll(" ", "");

Is there a good reason for this ?
As written before space can be a valid input.

Thanks

PR #2464, opened 2018-06-04
Remove replaceAll with trim

ctruzzi:
#### What's this PR do/fix?
Adds support for spaces in allowable Fields for 2.8.0 

Specifically Adds changes from 2.6.0 to 2.8.0 https://github.com/springfox/springfox/commit/fcf4a036fcfed4fea9e7859985af4bb5bfdf3682

We don't need both changes from #1211 since line 80 does the following `Iterable<String> split = Splitter.on(',').trimResults().omitEmptyStrings().split(trimmed);`

There were none prior

See #2446 
#2446


ctruzzi:
@dilipkrish Please review

dilipkrish:
@ctruzzi thanks! would you be able to add a test for this change?

ctruzzi:
@dilipkrish Yep can do

Issue #2448, opened 2018-05-28
X-Forwarded-prefix not work as expected since spring version 4.3.15


rainoko:
Springfox version 2.8.0

Problem appeared since spring version 4.3.15 or 5.0.5
With issue fix of https://jira.spring.io/browse/SPR-16650 

This issue modified ServletUriComponentsBuilder to not use context path as suffix of X-forwarded-prefix value. Instead now spring replaces path.

springfox prepends base path with x-forwarded-prefix in HostNameProvider

```
private static String prependForwardedPrefix(
      HttpServletRequest request,
      String path) {

    String prefix = request.getHeader("X-Forwarded-Prefix");
    if (prefix != null) {
      return prefix + path;
    } else {
      return path;
    }
  }
```

dilipkrish:
This solution needs to work correctly based on spring version it appears. Thanks for linking to the spring issue. Since you're experiencing the issue and have context, would love it if you could submit a PR to address this.

rainoko:
Ok I try to fix it. 
Should be ok to use org.springframework.core.SpringVersion.getVersion() to determine version and then adding or replacing path.
Is it ok?

dilipkrish:
I did not know that existed :), so yes that would be perfect

PR #2465, opened 2018-06-05
X-Forwarded-Prefix should replace basePath in newer spring versions

rainoko:
#### What's this PR do/fix?
since spring 4.3.15/5.0.5 x-forwarded-prefix replaces basePath

there are unit tests.
#2448


rainoko:
It seems that codecov is littlebit too strict for my work. Actually there wasn't any unittest for HostNameProvider. I added tests for version specific part. But all other code is just as it was before. So I didnt do test there.

Also it seems that version check code is actually covered but codecov thinks that some part of it is not.

dilipkrish:
@rainoko thanks for the PR! Will merge it in shortly! 🤘

dilipkrish:
@rainoko is there a reason why the spring version is restricted to 5.0.5 and not just be set to greater than 5.0.0?

rainoko:
According https://jira.spring.io/browse/SPR-16650

Fix versions 4.3.15, 5.0.5

dilipkrish:
Yeah I noticed that after messaging you :) thanks for your PR!

Issue #2511, opened 2018-06-25
Since version 2.9.0, APIs with @PathVariable are not picked up anymore due to Spring 4.3.3 code


dbaje:
In our project we are currently using spring 4.2.X and we recently upgraded to version 2.9.0 of this library. After upgrading we discovered that some of our APIs didn't appear anymore. After some digging, we discovered that an error was thrown and then caught which made troubleshooting difficult.

The issue is in the [ParameterRequiredReader]( https://github.com/springfox/springfox/blob/master/springfox-spring-web/src/main/java/springfox/documentation/spring/web/readers/parameter/ParameterRequiredReader.java) class

It's using fields in the PathVariable annotation that was introduced in Spring 4.3.3. Using a lower Spring version triggers a "MethodNotFound" exception. In our case, upgrading Spring is not an option.

As a work around, we have extended the offending class, modified the code. Then we told Spring to use ours as Primary. So we got it currently working.

There's a way for the code to do the same thing but without breaking compatibility

This:
```
Optional<PathVariable> pathVariable = methodParameter.findAnnotation(PathVariable.class);
    if (pathVariable.isPresent()) {
      String paramName = MoreObjects.firstNonNull(
          emptyToNull(pathVariable.get().name()),
          methodParameter.defaultName().orNull());
      if (pathVariable.get().required() ||
          optionalButPresentInThePath(operationContext, pathVariable.get(), paramName)) {
        requiredSet.add(true);
      }
    }
```

Could become this:

```
Optional<PathVariable> pathVariable = methodParameter.findAnnotation(PathVariable.class);
    if (pathVariable.isPresent()) {
      Map<String, Object> attributes = AnnotationUtils.getAnnotationAttributes(pathVariable.get());
      String paramName = MoreObjects.firstNonNull(
          emptyToNull((String) attributes.get("name")),
          methodParameter.defaultName().orNull());

      boolean required = !attributes.containsKey("required") || (Boolean) attributes.get("required");
      if (required ||
          optionalButPresentInThePath(operationContext, paramName)) {
        requiredSet.add(true);
      }
    }
```

Then this:
```
private boolean optionalButPresentInThePath(
      OperationContext operationContext,
      PathVariable pathVariable,
      String paramName) {

    return !pathVariable.required()
         && operationContext.requestMappingPattern().contains("{" + paramName + "}");
  }
```

To This:
```
private boolean optionalButPresentInThePath(
          OperationContext operationContext,
          String paramName) {

[...]

PR #2513, opened 2018-06-25
Bug/2511

dbaje:
#### What's this PR do/fix?
Bug #2511
No tests as this is to slove backwards compatibility
Attribute "name" and "required" in PathVariable annotation were added in Spring 4.3.3. Using the version of the code causes an exception.
#2511


dilipkrish:
@dbaje thanks for the pr!!

dbaje:
Made some changes and used the [spring capabilities utility](https://github.com/springfox/springfox/blob/master/springfox-swagger-common/src/main/java/springfox/documentation/swagger/common/SpringVersionCapability.java). I needed to move it so it would be available for the rest of the code. Also made some changes to be compatible with mocking and version checking in tests.

Issue #1570, opened 2016-11-16
Support for Response examples


KinOSurvival:
Any plans to support http://swagger.io/specification/#exampleObject?

I trace down to ResponseMessage.java, and found out that it doesn't have "example" property.

This close down my own plugin, trying to insert response samples(after parsing my own annotations)

I would love to have some suggestions on this. Thanks a lot.

dilipkrish:
Its simple enough to add it. Would you like to try your had at submitting a PR?

KinOSurvival:
yes, just one quick question:
@ApiResponse in swagger-annotation.jar is used to donate the Swagger Response, but it doesn't include examples right now. I guess we have to use our own annotation, and which annotation should I use?

dilipkrish:
@KinOSurvival gaah! that can be a problem. You could mint your own @Example annotation.

blvp:
there is some pitfalls with original annotation it is not required to set media type for example property. 
```java
@RequestMapping(consumes="application/json")
public void doSmth(
    @ApiParam(examples=@Example({@ExampleProperty("{\"test\":\"val\"}")})) TestVal testVal
) {}
```
Ignore it or extract Consumes from OperationContext?
And what to do when we have several ExampleProperty without mediaType or we have one ExampleProperty with several consumes? Like so:
```java
@RequestMapping(consumes={"application/json", "application/xml"})
public void doSmth(
    @ApiParam(examples=@Example({@ExampleProperty("{\"test\":\"val\"}")})) TestVal testVal
) {}
```
or 
```java
@RequestMapping(consumes={"application/json")
public void doSmth(
    @ApiParam(examples=@Example({
                @ExampleProperty("{\"test\":\"val\"}"),
                @ExampleProperty("<test>val</test>")
           })
    ) 
            TestVal testVal
) {}
```
I know that last  2 cases are least possible, but we must define some default logic for it like skipping or try to parse examples with defined consumes. 
By default we know if we have one consumes and one example it surely with consumes media type.

dilipkrish:
@blvp Yeah I agree. My personal opinion is that the swagger annotations are the ***manual knobs*** and tweaks to get the specification just right. They are a last resort for when the spring annotations do not describe enough for e.g. descriptions etc. So at that point Im cool with experimentation on the consumers part to set the right media types on the example.

quickwind:
When will this be added?

dilipkrish:
[...]

PR #2515, opened 2018-06-26
#1570 Add examples support

jasouyris:
Support ApiResponse examples.

**Summary**
Transforms @Example.value() (i.e. @ExampleProperty[]) into a List\<Example\> that then will be transformed into a Map<String, Object> that then will be used to populate the examples map in io.swagger.models.Response

**Details**
1) Added a List\<Example\> to ResponseMessage
2) Added a small block of code in SwaggerResponseMessageReader (where an ApiResponse is transformed into a ResponseMessage) that will transform an ExampleProperty into an Example and then add it to the ResponseMessage using the field above.
3) Transform the List\<Example\> to a Map<String, Object> in ServiceModelToSwagger2Mapper, and populate the Response.examples with it.

**Test**
1) Added additional UT to ResponseMessageSpec, ResponseMessageBuilderSpec and ServiceModelToSwagger2MapperSpec to cover newly added functionality.
2) Added a ResponseExampleTestController that will be used by SwaggerResponseMessageReaderSpec to test examples support.
3) Added UT for new ExamplesMapper class.
4) Added a new endpoint in FeatureDemonstrationService (not sure if this is the right place), to test the contract using the contract tests.

### DIFF
--- a/springfox-core/src/main/java/springfox/documentation/builders/ResponseMessageBuilder.java
+++ b/springfox-core/src/main/java/springfox/documentation/builders/ResponseMessageBuilder.java
@@ -19,6 +19,7 @@
 
 package springfox.documentation.builders;
 
+import springfox.documentation.schema.Example;
 import springfox.documentation.schema.ModelReference;
 import springfox.documentation.service.Header;
 import springfox.documentation.service.ResponseMessage;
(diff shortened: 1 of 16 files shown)


jasouyris:
Regarding the "Document empty method body" issues, what'd be the preferred way to solve it?

I could add a comment to the body, but I'm wondering why it doesn't complain for ResponseHeaderTestController... I guess it's due to it being an old class.

jasouyris:
Created a new PR [here](https://github.com/springfox/springfox/pull/2530) with the latest code from master plus fixed all the codacy issues.

Issue #2294, opened 2018-03-12
Return type Map<String, Set<String>> using Callable is not generated correctly by springfox


deepcabinwala:
**Java Controller Method**

    @ApiOperation(value ="To get device status", response = DeviceStatus.class, responseContainer = "Map")
    @GetMapping("/status")
    public Map<DeviceStatus, Set<UUID>> getSummaryDeviceStatusGroupedByStatus() {
        return deviceService.getAllDeviceStatusByStatus();
    }

Swagger Generated JSON:
"Map«string,Set«string»»": {
			"type": "object",
			"additionalProperties": {
				"$ref": "#/definitions/Set"
			}
		},

**Error**:
Resolver error at definitions.Map«string,List«object»».additionalProperties.$ref
Could not resolve reference because of: Could not resolve pointer: /definitions/List does not exist in document.

dilipkrish:
This seems like a duplicate of #2274. What version are you using?

deepcabinwala:
I am using springfox v2.8.0. When I don't use callable than it works fine without an error but adding callable throws the error. In the issue, you referred doesn't say anything about callable.

dilipkrish:
You need to provide an alternate type rule for callable  that converts `Callable<T>` =>`T`

Issue #2274, opened 2018-03-01
springfox cannot translate Map<String, List<SomeObject>> to Swagger spec correctly


europa1613:
For me the spec for Map<String, List<SomeObject>> data; is generated incorrectly by springfox like below. And the List has no definition. It does add the definition to SomeObject.

          `"data": {
                "type": "object",
                "additionalProperties": {
                    "$ref": "#/definitions/List"
                }
            }`
Versus the correct translation:

           `"data": {

                "type": "object",

                "additionalProperties": {

                    "type": "array",

                    "items": {

                        "$ref": "#/definitions/SomeObject"

                    }

                }

            }`
I'm using springfox-swagger2 annd springfox-swagger-ui versions 2.4.0. Also tried using 2.8.0 versions. 
Also tried annotating the property:
@ApiModelProperty(dataType="Map[String, java.util.List[SomeObject]]", reference="com.abc.SomeObject")

Any help/pointer is much appreciated.
Thanks!

MaksimOrlov:
Hi, @europa1613.
Could you, please, share the source code of the related controller?
And may be your `List` is not parameterized?

MaksimOrlov:
`@ApiModelProperty(dataType = **SomeType**)` leeds to calling ```class.forName()```, and requires a class name without parameters.

europa1613:
@MaksimOrlov Thanks for the reply. I did have it parameterized, was a typo in the question.

Unfortunately, I cannot share the actual code. But I am able to reproduce the issue within my public repository. Here is the link to the controller. 

https://github.com/europa1613/spring-boot/blob/master/rest-web-services/src/main/java/com/rest/services/restwebservices/customer/CustomerController.java

All the models are contained within the same package as Controller. The specific property not coming thru correctly is Customer --> AccountDetailsData(extends Data) --> **private Map<String, List<AccountsData>> accountsData** (in Data);

Please let me know your analysis. 

Thanks so much!

MaksimOrlov:
I've cloned your project and ran it with your configuration, received that definitions

```
"definitions": {
		"Account": {
			"type": "object",
			"properties": {
				"accountNumber": {
					"type": "string"
				},
				"accountType": {
					"type": "string"
				}
			}
		},
		"Payment": {
			"type": "object",
			"properties": {
				"paymentAmount": {
					"type": "string"
				}
			}
		},
		"AccountDetailsData": {
			"type": "object",
			"properties": {
				"accountsData": {
					"type": "object",
[...]

Issue #1881, opened 2017-06-22
Model for Map<String, List<Something>>


smwurster:
I've got an endpoint that returns a `Map<String, List<Something>>`.  This comes out as an empty (`{}`) model in the Swagger UI.  The `Something` class is annotated correctly, and I can see its reference definition in the resulting JSON, however nothing is referencing that definition.

Is there a way I can get this response type to come out correctly in the JSON / UI?  I've tried a couple Alternate Type Rules but nothing seemed to work.

Thanks.

dilipkrish:
What does the api-docs look like?

smwurster:
The api-docs does indeed show the reference to the definition in the endpoint's schema:

`"schema":{"type":"object","additionalProperties":{"type":"array","items":{"$ref":"#/definitions/Something"}}}`

However the UI shows it as `inline_model_9 {}`

The definition for `Something` in the api-docs looks to be a typical definition.

dilipkrish:
The generated spec seems correct. The swagger-ui is probably to blame.

smwurster:
I think you're right.  The current online Swagger editor seems to show everything correctly, so it must just be an issue with the version currently being packaged.  That's fine for now.

gihad:
I hit the exact same issue today - When I paste the api-docs in http://editor.swagger.io/ everything looks good. Is there a way to fix this for the embedded api-docs ui?

dilipkrish:
We could use some help upgrading to 3.0 for sure 🤔

europa1613:
@smwurster For me the spec for `Map<String, List<Something>> data;` is generated incorrectly by springfox like below.

              `"data": {
                    "type": "object",
                    "additionalProperties": {
                        "$ref": "#/definitions/List"
                    }
                }`
Versus the correct translation:
             `"data": {
                    "type": "object",
                    "additionalProperties": {
                        "type": "array",
                        "items": {
                            "$ref": "#/definitions/Something"
                        }
                    }
                }`

I'm using springfox-swagger2 annd springfox-swagger-ui versions 2.4.0. 

Any help/pointer is much appreciated.
Thanks!

dilipkrish:
@europa1613 That is working as expected in 2.8.0. Please see test above ☝️

europa1613:
I opened #2274. Yes its working in 2.8.0. Thanks @dilipkrish

vpm911:
[...]

PR #2544, opened 2018-07-08
Fix #2539

### LINKED ISSUE #2539: Generic response model with Map<String, List<T>>
What version of the library are you using?
**2.8.0, and I tried 2.9.2 with no luck.**

What kind of issue is this?
The issue is much more like #2274 (also #1881 , #2294 ), but I used a generic response model.
The code looks like below:

```java
public class ResponseData<T> {
    private int code = 200;

    @JsonInclude(JsonInclude.Include.NON_NULL)
    private String message;

    @JsonInclude(JsonInclude.Include.NON_NULL)
    private T data;
}
```

```java
@ApiOperation(value = "Get people group by city")
@GetMapping("/test")
public ResponseData<Map<String, List<People>>> testMap() {
    return ResponseData.success(Collections.singletonMap("New York", Collections.singletonList(new People("John Doe", 20, "New York"))));
}
```

I've created a demo repo with spring boot: [https://github.com/draco1023/spec-issue-demo](https://github.com/draco1023/spec-issue-demo)

Start the demo then open http://127.0.0.1:8080/v2/api-docs, there will be a definition in definitions section, and the List has no definition.
```json
"Map«string,List«People»»":
        {
            "type": "object",
            "title": "Map«string,List«People»»",
            "additionalProperties":
            {
                "$ref": "#/definitions/List"
            }
        }
```

It is interesting to note the definition of ResponseData<Map<String, List<People>>> is right.
```json
"ResponseData«Map«string,List«People»»»":
        {
            "type": "object",
            "properties":
            {
                "code":
                {
                    "type": "integer",
                    "format": "int32"
                },
                "data":
                {
                    "type": "object",
                    "additionalProperties":
                    {
                        "type": "array",
                        "items":
                        {
                            "$ref": "#/definitions/People"
                        }
                    }
                },
                "message":
                {
                    "type": "string"
                }
            },
            "title": "ResponseData«Map«string,List«People»»»"
        }
```

draco1023:
#### What's this PR do/fix?
Issue #2539 
No unit tests, but I've tested with the issue demo repo.
Maybe #2274 , #1881 , #2294

### DIFF
--- a/springfox-swagger2/src/main/java/springfox/documentation/swagger2/mappers/ModelMapper.java
+++ b/springfox-swagger2/src/main/java/springfox/documentation/swagger2/mappers/ModelMapper.java
@@ -53,6 +53,7 @@ import static java.util.Optional.*;
 import static java.util.stream.Collectors.*;
 import static springfox.documentation.schema.Collections.*;
 import static springfox.documentation.schema.Maps.*;
+import static springfox.documentation.schema.Types.*;
 import static springfox.documentation.swagger2.mappers.EnumMapper.*;
 import static springfox.documentation.swagger2.mappers.Properties.*;
 
@@ -125,6 +126,8 @@ public abstract class ModelMapper {
       return new ArrayProperty(resolvedTypeToProperty(collectionElementType(type)));
     } else if (isMapType(type)) {
       return new MapProperty(resolvedTypeToProperty(mapValueType(type)));
+    } else if (isBaseType(type)) {
+      return property(typeNameFor(type.getErasedType()));
     } else {
       return new RefProperty(type.getErasedType().getSimpleName());
     }
--- a/springfox-swagger2/src/main/java/springfox/documentation/swagger2/mappers/ModelMapper.java
+++ b/springfox-swagger2/src/main/java/springfox/documentation/swagger2/mappers/ModelMapper.java
@@ -28,8 +28,8 @@ import io.swagger.models.Xml;
 import io.swagger.models.properties.AbstractNumericProperty;
 import io.swagger.models.properties.ArrayProperty;
 import io.swagger.models.properties.MapProperty;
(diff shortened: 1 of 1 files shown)


dilipkrish:
@draco1023 thanks for the PR! However the existing tests don't seem to be passing. Perhaps add a test of your own as well and describe what your problem was and how this fixes the problem

dilipkrish:
@draco1023 🎗

draco1023:
@dilipkrish Sorry, I don't have much time recently.
My PR always goes deep when Map and its subtype is met, but the schema perfers $ref, which caused ci tests failed.

dilipkrish:
@draco1023 not a problem. In general, even if the aren't any new tests, its should minimally pass the ci build to get pulled in.

Also there is no UI that will support this as the spec itself doesnt support this scenario.

draco1023:
image
image

@dilipkrish The figures above show the main differences between [swagger2/declaration-bugs-service.json](https://github.com/springfox/springfox/blob/master/swagger-contract-tests/src/test/resources/contract/swagger2/declaration-bugs-service.json) (left side) and the result using the PR (right side).
Since [ApplicationLang](https://github.com/springfox/springfox/blob/master/springfox-schema/src/test/java/springfox/documentation/schema/ApplicationLang.java) and [Language](https://github.com/springfox/springfox/blob/master/springfox-schema/src/test/java/springfox/documentation/schema/Language.java) both implement the Map interface, and in my opinion, there is no reason to distinguish between Map's (or Collection's) implemention class and interface.

dilipkrish:
@draco1023 thats fine, but couple of things
- Regardless, the test need to pass, so arguably you could make the change to the contract files but I dont believe this is the right fix.
- What if you had nested one level deeper? It doesnt seem correct that you'd inline all the models. The way that it renders currently, it preserves the types for e.g. Language.

Issue #1359, opened 2016-06-24
Unable to build on gradle or eclipse


mrethers:
using Gradle 2.2.1 I get the following eror:

`To honour the JVM settings for this build a new JVM will be forked. Please consider using the daemon: http://gradle.org/docs/2.2.1/userguide/gradle_daemon.html.
:buildSrc:clean UP-TO-DATE
:buildSrc:compileJava UP-TO-DATE
:buildSrc:compileGroovy
:buildSrc:processResources
:buildSrc:classes
:buildSrc:jar
:buildSrc:assemble
:buildSrc:compileTestJava UP-TO-DATE
:buildSrc:compileTestGroovy
startup failed:
Could not instantiate global transform class org.spockframework.compiler.SpockTransform specified at jar:file:/Users/mrethers/.gradle/caches/modules-2/files-2.1/org.spockframework/spock-core/1.0-groovy-2.4/ceaa8b69f274ed3de24da3e6a6c86f673b426d1a/spock-core-1.0-groovy-2.4.jar!/META-INF/services/org.codehaus.groovy.transform.ASTTransformation  because of exception org.spockframework.util.IncompatibleGroovyVersionException: The Spock compiler plugin cannot execute because Spock 1.0.0-groovy-2.4 is not compatible with Groovy 2.3.6. For more information, see http://versioninfo.spockframework.org
Spock artifact: file:/Users/mrethers/.gradle/caches/modules-2/files-2.1/org.spockframework/spock-core/1.0-groovy-2.4/ceaa8b69f274ed3de24da3e6a6c86f673b426d1a/spock-core-1.0-groovy-2.4.jar
Groovy artifact: file:/usr/local/Cellar/gradle/2.2.1/libexec/lib/groovy-all-2.3.6.jar

1 error

:buildSrc:compileTestGroovy FAILED

FAILURE: Build failed with an exception.

* What went wrong:
Execution failed for task ':compileTestGroovy'.
> Compilation failed; see the compiler error output for details.

* Try:
Run with --stacktrace option to get the stack trace. Run with --info or --debug option to get more log output.

BUILD FAILED

Total time: 5.511 secs
`

On eclipse using STS Gradle 3.7.2 I get `:buildSrc:compileJava UP-TO-DATE
:buildSrc:compileGroovy UP-TO-DATE
:buildSrc:processResources UP-TO-DATE
:buildSrc:classes UP-TO-DATE
:buildSrc:jar UP-TO-DATE
[...]

PR #2696, opened 2018-09-22
allow importing into eclipse

a1dutch:
This PR fixes the issue with not being able to import the project into eclipse using buildship (gradle integration).  This is detailed in #1359.

Eclipse is able to have multiple projects unlike intellij which can only have one(i assume intellij sets the working directory to the root of this folder), eclipse sets it working directory to its workspace which is a different directory to its project directory.  When it tries to find the last annotated tag via git its looking in the workspace directory which is not a git repo.

The fix simply adds the `-C <project root directory>` from the project so it can find the correct git repository.

I have imported the project into intellij and eclipse, all existing tests pass when run from a terminal.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -18,3 +18,5 @@ out/
 *.hprof
 node_modules/
 package-lock.json
+.eclipse/
+bin/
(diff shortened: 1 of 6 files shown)


dilipkrish:

ashirley:
I can confirm this fixed importing the project into eclipse (using the Spring Tools Suite 4 build of eclipse)

a1dutch:
@dilipkrish any plans to get this merged?

dilipkrish:
@a1dutch sorry for the really late reply. Im just coming out of a hiatus. I had a question about this change.

While this apparently doesn't have any problems, Im curious why the build system needs to know about how the IDE works?

a1dutch:
@dilipkrish the issue is to do with how the ide’s set different working directories. 

IntelliJ can only open a single project at a time and obviously sets the working directory to that project.

Eclipse on the other hand can import multiple projects into its workspace. The workspace is the working directory but not the directory of the project.

Gradle is aware of the absolute directory of the project it just needed to tell git what directory to use as the working directory is different

PR #2697, opened 2018-09-22
handle any extension type

a1dutch:
This PR allows any object to be added as a vendor extension just like the ListVendorExtension.

Currently you have to override the existing VendorExtensionsMapper in order to add non String type.

Test has been updated to include a check for an pojo type, iv'e also manually checked the correct values are present in the json.

### DIFF
--- a/springfox-swagger2/src/main/java/springfox/documentation/swagger2/mappers/VendorExtensionsMapper.java
+++ b/springfox-swagger2/src/main/java/springfox/documentation/swagger2/mappers/VendorExtensionsMapper.java
@@ -20,7 +20,6 @@ package springfox.documentation.swagger2.mappers;
 
 
 import org.mapstruct.Mapper;
-import springfox.documentation.service.ListVendorExtension;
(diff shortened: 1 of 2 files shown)


dilipkrish:

One observation is that it already supports `ObjectVendorExtension` is that not working for you?

a1dutch:
@dilipkrish the problem with `ObjectVendorExtension` is it acts more like a map than a key and a value.

e.g.

```
object: { key: value}
```

rather than

```
key: value
```

Its now also possible to have any other type, as it will map the correct types from the object to the json.  e.g. this is one of the vendor extensions i use:

```java
    public class ObjectExtension<T> implements VendorExtension<T> {

        private String name;
        private T value;

        public ObjectExtension(String name, T value) {
            this.name = name;
            this.value = value;
        }

        @Override
        public String getName() {
            return name;
        }

        @Override
        public T getValue() {
            return value;
        }
    }
```

a1dutch:
@dilipkrish any plans to get this merged?

Issue #2578, opened 2018-07-26
Optional ACAHs: X-XSRF-TOKEN 


iutipikin:
Hi everyone!

I'm using latest version (2.9.2) of this awesome project to describe my endpoints. Some of them are secured by Keycloak server (OAuth2 authorization flow) and i implemented a Swagger-UI OAuth2 security schema. Everything seems ok, but what's the problem? Keycloak server (which i'm not in control) not allow 'Access-Control-Allowed-Headers : x-xsrf-token' header, so then i'm trying to get JWT, request which made by swagger-ui always blocked by mu IDP. 

Can this feature - sending CSRF token to external IDP endpoint may be optional?

iutipikin:
Ok, I found what if I implement /csrf endpoint (for not getting HTTP/404) X-XSRF-TOKEN will be added to access headers automaticaly by this function **getCsrfFromCookie()** from _springfox-swagger-ui/src/web/js/csrf.js_. 
Is it possible to provide some flag from SecurityConfiguration to that function's context?

dilipkrish:
@iutipikin seems like you're further along in the research than I am. Would be happy to accept any fix you think might fix it.

olOwOlo:
Ok, I wrote it in #2434 . ORZ 😢 
Because I didn't know how to add options, I can only try to fetch the CSRF token. There are three steps:

1. fetch `http://your-base-url/`
2. fetch `http://your-base-url/csrf`
3. search your cookie

If your server do not support csrf, you'll see two 404 requests, but everything is ok and there are no errors. These requests are just trying to find your csrf token. If a csrf token is found, it will be automatically added to your request header.

I guess I found how to add options, I will send a pr to make these behaviors optional.

---

Relate issues: #2633 #2603 #2573

fennekit:
Pull request https://github.com/springfox/springfox/pull/2706

dilipkrish:
Thank you @fennekit

ajmalch:
> Thank you @fennekit

@dilipkrish , We are also facing the same issue #2633. Do you  think the PR @fennekit  created will resolve the issue and if yes, can we expect this in the next version?

PR #2706, opened 2018-09-30
Optional ACAHs: X-XSRF-TOKEN

fennekit:
#### What's this PR do/fix?
[#2578](https://github.com/springfox/springfox/issues/2578). Allow to exclude urls from the CSRF header
yes. For checking the configuration. No 
Like #2578 I ran into issues using KeyCloak with SpringFox. It is now possible to enable/disable CSRF support and/or filter out the header for certain domains.
CSRF headers are send for all requests.


dilipkrish:

PR #2721, opened 2018-10-06
Sort parameters by their order before sorting them by their name

shartte:
Rather than sorting parameters only by their name, sort them by their existing order field first, then use the name as a tiebreaker.
Parameter already implements Ordered, so this seems like a natural way of sorting parameters.
This also enables plugins to apply any custom parameter order they want, while the current sorting prevents that.
Currently parameters all have the initial order value of Integer.MIN_VALUE, so this change should not affect the current ordering without any additional plugins applied.

### DIFF
--- a/springfox-core/src/main/java/springfox/documentation/service/Operation.java
+++ b/springfox-core/src/main/java/springfox/documentation/service/Operation.java
@@ -81,7 +81,7 @@ public class Operation {
     this.isHidden = isHidden;
     this.securityReferences = toAuthorizationsMap(securityReferences);
     this.parameters = parameters.stream()
-        .sorted(byParameterName()).collect(toList());
+        .sorted(byOrder().thenComparing(byParameterName())).collect(toList());
     this.responseMessages = responseMessages;
     this.deprecated = deprecated;
     this.vendorExtensions = new ArrayList<>(vendorExtensions);
@@ -154,6 +154,10 @@ public class Operation {
 
   public List<VendorExtension> getVendorExtensions() {
     return vendorExtensions;
+  }
+
+  private Comparator<Parameter> byOrder() {
+    return Comparator.comparingInt(Parameter::getOrder);
   }
 
   private Comparator<Parameter> byParameterName() {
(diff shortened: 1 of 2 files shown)


mikecollard:
+1

raderio:
+1

ratoaq2:
+1

I was able to keep the parameters order just adding this plugin alongside the change of this PR

```
import org.springframework.core.Ordered;
import org.springframework.core.annotation.Order;
import org.springframework.stereotype.Component;

import springfox.documentation.service.ResolvedMethodParameter;
import springfox.documentation.spi.DocumentationType;
import springfox.documentation.spi.service.ParameterBuilderPlugin;
import springfox.documentation.spi.service.contexts.ParameterContext;
import springfox.documentation.swagger.common.SwaggerPluginSupport;

@Component
@Order(SwaggerPluginSupport.SWAGGER_PLUGIN_ORDER)
public class ParameterIndexOrderReader implements ParameterBuilderPlugin {

    private static final int PARAMETER_INITIAL_ORDER = Ordered.LOWEST_PRECEDENCE - 100;

    @Override
    public boolean supports(DocumentationType delimiter) {
        return true;
    }

    @Override
    public void apply(ParameterContext context) {
        ResolvedMethodParameter methodParameter = context.resolvedMethodParameter();
        context.parameterBuilder().order(PARAMETER_INITIAL_ORDER + methodParameter.getParameterIndex());
    }
}

```

ratoaq2:
The change in this PR doesn't affect the current 2.9.x behavior (alphabetically sorted)

But IMO, the parameter index should be kept and that should be the default behavior. The API spec is used to auto generate code and sorting it alphabetically by default doesn't seem reasonable.

ratoaq2:
As a side note, OperationBuilder, ParameterBuilder and similars could be spring components with scope 'prototype' so we could easily override them

tholu:
@shartte Can you resolve the conflict? I would also like to keep the defined order instead of sorting parameters alphabetically and am looking forward to see this landing in 3.0.

shartte:
Sorry, no, I won't put any more time in. It seems like this PR has no chance of being merged given how long it is open.

dilipkrish:
@shartte sorry it took so long. I've resolved the conflicts and merged.

dilipkrish:
Thanks for the PR! 🤘

antechrestos:
Hi,

@dilipkrish Will it be released soon? This change is quite usefull INMO.

dilipkrish:
@antechrestos yes Im trying to as soon as I can... see #3070

schrek1:
looks still broken

PR #2722, opened 2018-10-06
Create a FormParameter model object instead of BodyParameter

### LINKED ISSUE #2717: Primitive RequestPart parameters generate incorrect Swagger Model
Method parameters of the following form will generate incorrect parameters in Swagger 2. Non-validating parsers will not immediately choke on it, but code generators like swagger-codegen or openapi-codegen will.

```java
  public SomeReturnValue webMethod(
    @RequestPart("file") MultipartFile file,
    @RequestPart("description") String description) {
...
```

The generated spec will look like this:
```json
                    {
                        "name": "file",
                        "in": "formData",
                        "description": "file",
                        "required": true,
                        "type": "file"
                    },
                    {
                        "in": "formData",
                        "name": "description",
                        "description": "description",
                        "required": true,
                        "schema": {
                            "type": "string"
                        }
                    }
```

This is a problem because parameters with ``in=formData`` can't have a schema, only a type is allowed. The spec doesn't allow formData parameters to be anything other than a primitive.

The expected JSON should be:
```json
                    {
                        "name": "file",
                        "in": "formData",
                        "description": "file",
                        "required": true,
                        "type": "file"
                    },
                    {
                        "in": "formData",
                        "name": "description",
                        "description": "description",
                        "required": true,
                        "type": "string"
                    }
```

shartte:
This fixes  #2717 by creating the right Swagger model class for parameters of type "formData".
Since those parameters only support a single example, I've also opted to simply copy over the first one.

### DIFF
--- a/springfox-swagger2/src/main/java/springfox/documentation/swagger2/mappers/ParameterMapper.java
+++ b/springfox-swagger2/src/main/java/springfox/documentation/swagger2/mappers/ParameterMapper.java
@@ -36,13 +36,27 @@ import springfox.documentation.schema.ModelReference;
 import java.util.List;
 import java.util.Map.Entry;
 import java.util.Optional;
+import java.util.Set;
+import java.util.stream.Stream;
 
-import static springfox.documentation.schema.Types.*;
-import static springfox.documentation.swagger2.mappers.EnumMapper.*;
-import static springfox.documentation.swagger2.mappers.Properties.*;
+import static java.util.stream.Collectors.toSet;
+import static springfox.documentation.schema.Types.isBaseType;
+import static springfox.documentation.swagger2.mappers.EnumMapper.maybeAddAllowableValues;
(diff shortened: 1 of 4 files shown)


shartte:
I've adapted the class so that it only maps to FormParameter now if the type is one of the types supported by the spec. In other cases it falls back to BodyParameter, which is seemingly wrong, but matches previous behavior better.
I've also added a few unit tests for these cases.
Since the "spec" tests don't run on my own machine, I hope my adapted tests data will work on the CI.

Carmageddon:
Whats up with this PR? Can you please get it fixed so we would not have to build based on your repo on Monday :)

dilipkrish:
@shartte thanks for the PR. It appears that the contract tests are failing. Would you mind taking a look?

kremers:
Having issues with #2717
Very happy with the solution, hope it makes it in 3.0.0-SNAPSHOT soon

PR #2725, opened 2018-10-07
Ignore the base class in the list of @JsonSubTypes

### LINKED ISSUE #2724: Polymorphism can lead to a recursive model
I've noticed this by the virtue of StackOverflowExceptions when I generate code from my Swagger spec generated by 2.9.

````java
@JsonTypeInfo(use = JsonTypeInfo.Id.NAME)
@JsonSubTypes([
  @JsonSubTypes.Type(value = A, name = "a"),
  @JsonSubTypes.Type(value = B, name = "b")
])
class A {
}
class B {
}
```

This generates a schema that'll look like the following:
``` 
    "A": {
      "title": "A",
      "allOf": [
        {
          "$ref": "#/definitions/A"
        },
        {
          "type": "object",
...
``` 

So effectively a self-recursive data structure.

shartte:
Fixes #2724 by ignoring the base class found in the `@JsonSubTypes` annotation.

The generated schema still has issues, but at least it doesn't crash code generators anymore.


dilipkrish:
Thanks for the PR, will review and pull in shortly @shartte

PR #2745, opened 2018-10-16
Expose GenericTypeNamingStrategy in the Docket

PissedCapslock:
The current `forCodeGeneration` method only allows to switch between two hardcoded  `GenericTypeNamingStrategy` implementations. Neither of them can deal with both generics and spaces. E.g. if you have a class

````
@ApiModel(value = "Foo Bar", description = "lorem ipsum")
public final class FooBar<T>{

}
````
the generated Swagger specification file still results in  `$ref values must be RFC3986-compliant percent-encoded URIs` warnings, even when calling `forCodeGeneration(true)`.

By allowing to replace the `GenericTypeNamingStrategy`, people can set their own naming strategy

### DIFF
--- a/springfox-spring-web/src/main/java/springfox/documentation/spring/web/plugins/Docket.java
+++ b/springfox-spring-web/src/main/java/springfox/documentation/spring/web/plugins/Docket.java
@@ -364,11 +364,25 @@ public class Docket implements DocumentationPlugin {
    *
    * @param forCodeGen - true|false determines the naming strategy used
    * @return this Docket
+   * @see #genericTypeNamingStrategy
    */
   public Docket forCodeGeneration(boolean forCodeGen) {
     if (forCodeGen) {
       genericsNamingStrategy = new CodeGenGenericTypeNamingStrategy();
     }
+    return this;
+  }
+  
+  /**
+   * Replaces the springfox.documentation.spi.schema.GenericTypeNamingStrategy.
+   * This allows to replace the influence the strategy to provide names for generic types in the swagger output.
+   *
+   * @param strategy - the new strategy
+   * @return this Docket
+   * @see #forCodeGeneration
+   */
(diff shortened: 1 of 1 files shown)


dilipkrish:
Im still wondering when you'd have a case with a space in the name, that is valid for a type? Is this a real problem you're facing? Also the % character issue, is a [Swagger-ui](https://github.com/swagger-api/swagger-ui) issue.

PissedCapslock:
In our project, we generate the swagger specification file based on annotations in the code. From that swagger specification file, we generate 

* documentation (using springfox)
* Swagger UI console
* Probably client SDKs in the future

We have return types with generics: `Envelope<FooBar>` and `Envelop<Bar>`. Sometimes the return type is also a collection: `Envelope<List<FooBar>>`.

Because `FooBar` is just a class name and not correct English, that class is annotated with

````
@ApiModel(value = "Foo bar", description = "...")
````
The result of this is an invalid Swagger specification file.

Using the `Docket#forCodeGeneration` method, I can generate a Swagger specification file which is valid, but the definitions look ugly to a human. E.g. `EnvelopeOfListOfFooBar`. So now I have a valid file, but ugly documentation.

When I logged this issue, I was hoping to get both: valid spec file and readable documentation.
I was hoping to override this naming strategy and provide human readable names.
But I think this PR is not sufficient to fix my problem.

What I think I actually need is a way to specify the `title` property of a [definition object](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/2.0.md#definitionsObject) independently.

More concrete

````
{
  "FooBar": {
    "type": "object",
    "properties": { },
    "title": "Foo bar"
  }
````

I think I need something on the `@ApiModel` which allows me to specify the `title` ("Foo bar") independent from the `value` ("FooBar"). As far as I could tell, they now are always the same meaning your title is suddenly subject to the same validation rules.
That way I can use a value which is valid according to the spec, and use a human readable title.

PR #2747, opened 2018-10-18
Keep query params on redirect

viktorgunnarson:
Keep the query params when redirecting Swagger-UI.

#### What's this PR do/fix?
Updates a example in the documentation relating to moving swagger-ui. The updated example ensures that the query parameters are kept when redirecting.

-
The current example hard codes a query parameter that causes issues.

-

### DIFF
--- a/docs/asciidoc/common-problems.adoc
+++ b/docs/asciidoc/common-problems.adoc
@@ -259,7 +259,7 @@ For e.g. One could move Swagger UI under `/documentation` using this code.
 	@Override
 	public void addViewControllers(ViewControllerRegistry registry) {
 	
-		registry.addRedirectViewController("/documentation/v2/api-docs", "/v2/api-docs?group=restful-api");
+		registry.addRedirectViewController("/documentation/v2/api-docs", "/v2/api-docs").setKeepQueryParams(true);
         	registry.addRedirectViewController("/documentation/swagger-resources/configuration/ui","/swagger-resources/configuration/ui");
         	registry.addRedirectViewController("/documentation/swagger-resources/configuration/security","/swagger-resources/configuration/security");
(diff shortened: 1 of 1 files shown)


viktorgunnarson:
Ping @dilipkrish

docs/asciidoc/common-problems.adoc
```diff
@@ -259,7 +259,7 @@ For e.g. One could move Swagger UI under `/documentation` using this code.
 	@Override
 	public void addViewControllers(ViewControllerRegistry registry) {
 	
-		registry.addRedirectViewController("/documentation/v2/api-docs", "/v2/api-docs?group=restful-api");
+		registry.addRedirectViewController("/documentation/v2/api-docs", "/v2/api-docs").setKeepQueryParams(true);
```
dilipkrish: Apply it to all the redirects?

```suggestion
		registry.addRedirectViewController(
				"/documentation/v2/api-docs", 
				"/v2/api-docs")
			.setKeepQueryParams(true);
        	registry.addRedirectViewController(
				"/documentation/swagger-resources/configuration/ui",
				"/swagger-resources/configuration/ui")
			.setKeepQueryParams(true);
        	registry.addRedirectViewController(
				"/documentation/swagger-resources/configuration/security",
				"/swagger-resources/configuration/security")
			.setKeepQueryParams(true);
        	registry.addRedirectViewController(
				"/documentation/swagger-resources", 
				"/swagger-resources")
			.setKeepQueryParams(true);
```

viktorgunnarson: Sure, but I think that today only the first redirect uses query parameters.

dilipkrish: Oh thats right... thanks for pointing that out.

PR #2778, opened 2018-11-14
Preserve examples

dschulten:
#### What's this PR do/fix?
This PR ensures that Examples added with ApiResponse annotations are no longer dropped when merging ResponseMessages in OperationBuilder.

OperationBuilderSpec
def "Response message builder is non-destructive"

I need examples to add response example snippets created by spring-restdocs in spring-integration for springfox

None


dilipkrish:
Thank you @dschulten for catching that!

Issue #2817, opened 2018-12-12
Allow to set up the base path in the Docket configuration (Springfox 3.0.0-SNAPSHOT)


codependent:
- Version 3.0.0-SNAPSHOT
- Feature Request

I am using Springfox 3.0.0-SNAPSHOT in a Spring Boot 2.1.1 WebMvc project. It generates a documentation with basePath : "/" and paths like: /myapp/api/v1/customers. What I would like is basePath: "/myapp" and paths /api/v1/customers.

I have searched for information about how to customize the base path and the endpoints, found [this issue where jonaskoperdraat provides a solution](https://github.com/springfox/springfox/issues/1443) and tried it:

```
    @Bean
    public Docket api() {
        return new Docket(DocumentationType.SWAGGER_2)
        		.host("localhost:8081")
                .pathProvider(new PathProvider() {
                	
                   private final String ROOT = "/myapp";
                	
                    @Override
                    public String getOperationPath(String operationPath) {
                        return operationPath.replace(ROOT, "");
                    }
                    @Override
                    public String getResourceListingPath(String groupName, String apiDeclaration) {
                        return null;
                    }
                })
```
My new paths are ok: /api/v1/customers, but the base path keeps being "/". I noticed the suggestion of explicitly setting Docket().host() but it didn't work either.

I have debugged the source code and found this on startup `springfox.documentation.spring.web.scanners.ApiDocumentationScanner`:

```
public Documentation scan(DocumentationContext context) {
    ApiListingReferenceScanResult result = apiListingReferenceScanner.scan(context);
    ApiListingScanningContext listingContext = new ApiListingScanningContext(context,
        result.getResourceGroupRequestMappings());

    Map<String, List<ApiListing>> apiListings = apiListingScanner.scan(listingContext);
    Set<Tag> tags = toTags(apiListings);
    tags.addAll(context.getTags());
    DocumentationBuilder group = new DocumentationBuilder()
        .name(context.getGroupName())
        .apiListingsByResourceGroupName(apiListings)
        .produces(context.getProduces())
        .consumes(context.getConsumes())
        .host(context.getHost())
        .schemes(context.getProtocols())
        .basePath(ROOT)
        .extensions(context.getVendorExtentions())
        .tags(tags);
```
 As you can see there's an explicit `.base(ROOT)` invocation. Couldn't `DocumentationContext` offer a context.getBasePath() method that was set based on another `Docket().basePath()` operation?

If not, what's the way to customize this?

[...]

PR #2819, opened 2018-12-12
basePath configuration support in Docket

codependent:
This pull request allows the configuration of the Swagger basePath from the Docket, e.g:

```
    @Bean
    public Docket api() {
        return new Docket(DocumentationType.SWAGGER_2)
                .basePath("/myapp")
```

The project tests pass but there aren't additional tests included in this issue. To check it just generate a new project and add the basePath() info in the Docket. I have checked manually that not adding the basePath() generates "/" as before. 
Background context: there was no straightforward or working way to configure the Swagger basePath.
The associated issue is: https://github.com/springfox/springfox/issues/2817

### DIFF
--- a/springfox-spi/src/main/java/springfox/documentation/spi/service/contexts/DocumentationContext.java
+++ b/springfox-spi/src/main/java/springfox/documentation/spi/service/contexts/DocumentationContext.java
@@ -70,36 +70,38 @@ public class DocumentationContext {
   private Set<String> produces;
   private Set<String> consumes;
   private String host;
+  private String basePath;
   private Set<String> protocols;
   private boolean isUriTemplatesEnabled;
   private List<VendorExtension> vendorExtensions;
 
   public DocumentationContext(
-      DocumentationType documentationType,
-      List<RequestHandler> handlerMappings,
-      ApiInfo apiInfo, String groupName,
-      ApiSelector apiSelector,
-      Set<Class> ignorableParameterTypes,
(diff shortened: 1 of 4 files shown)


dilipkrish:
Thanks for the PR! This is solved using `PathProvider` implementation. What is the problem we're trying to solve here?

sourcekick:
@dilipkrish This solution would have been intuitive.

Where can I find documentation how to apply the `PathProvider`?

We are using springfox-swagger2 `2.9.2`

mightybruno:
@dilipkrish it's not the same thing. Sure, you can **add** stuff using `PathProvider`, but you can't **substract** it. Like let's say you have swagger docs under
`localhost/swagger/swagger-ui.html`
The base path would be equal to `/swagger` without any additional configuration. Now, you can make the resource path more specific (e.g.  `/swagger/foo`) with the help of `PathProvider` but from what I know, you can't make it more generic, that is just `/`.

To accomplish that I had to override `ObjectMapper` bean where if the passed argument is an instance of `Swagger` the basePath is set to `/`. 

If this PR were to be merged I wouldn't have to use such an unintuitive _hack_.

codependent:
@dilipkrish Could you reconsider merging this?

mightybruno:
@codependent there are conflicts tho, will you resolve them? also, there are quite a lot of formatting changes, would be nice to revert them (in order to preserve the projects conventions)

PR #2826, opened 2018-12-17
Only hidden operation when the hidden property is true

tobecodingmaster:
I have custom constraint check on the swagger specification. I need hidden the operation which violated the constraint; But it is not work on the parameter constraint check which extended `ParameterBuilderPlugin`; When deep into the source code of `springfox`, i found the `OperationHiddenReader` set hidden false. And it processing after the  `ParameterBuilderPlugin` lifecycle. So can only hidden operation when the hidden property is true? It is convenient for developer to do custom constraint check.


dilipkrish:
Is this simply solved by giving your custom `ParameterBuilderPlugin` a higher precedence?

tobecodingmaster:
> Is this simply solved by giving your custom `ParameterBuilderPlugin` a higher precedence?

No, the `ParameterBuilderPlugin` is processing in `OperationParameterReader`, which is higher precedence than `OperationHiddenReader`; So the OperationHiddenReader set hidden false can override the setting in the `ParameterBuilderPlugin`  ;

Issue #2064, opened 2017-10-09
java 9 library compatibility issue


amartinm7:
Library versions:

- JAVA 9

- springboot 1.5.3.RELEASE

- 'io.springfox:springfox-swagger2:2.7.0'

-  'io.springfox:springfox-swagger-ui:2.7.0'

-  gradle 2.13

 When I'm trying to use the Swagger2 jars in my "module-info.java" file I have the error: 

> - module com.example read package "springfox.documentation.service" from "springfox.core" and "springfox.spi" 

The problem as the error message is saying, the package with the same name is located is two different locations. So I think the solution can be or rename the packages in order to have different package names or unify all the classes of the two packages in only one package. And that is. 

Regards

dilipkrish:
The libraries are not java 9 compatible yet.

zigterman:
Are there any plans to fix this shortly?

dilipkrish:
@zigterman I don't know much about what it would take unfortunately. Would really appreciate someone taking a look at this

zigterman:
Hi,

we where also starting with java 9 and taking a look at jigsaw (the java 9 module system), so I'm not an expert as well. But I don't think there is much that have to be changed. @amartinm7 describe the problem already. 

There are 2 different springfox libraries that use the same package structure. This gives a error, because for jigsaw you have to define them in your module-info file. But as there are, jigsaw doesn't know which one to use.

There are 3 solutions I see:
1. rename one off the packages
2. merge the code off the same packages into 1 project
3. define in the projects a module-info.java class and export only the packages/classes that may be used in other projects.

The last one is the best option because then you are complete compatible with java 9, but I think option 1 or 2 is also needed than.

At this moment we are using java 9, but not the jigsaw module structure. Because this problems happens also with other libraries unfortunately :-(.

bennyflint:
This is becoming a very problematic issue. I don't see how anyone using Springfox can move to Java 9+ and modules without solving this. Java 8 support ends in January. Has anyone attempted a fix yet?

rgoers:
I notice you say Java 9 AND modules. Does it work with Java 9 without using modules? From my experience with trying to add module support to Log4j I can safely say you are going to have problems working with them.

bennyflint:
I don't yet know whether it works without modules, but @zigterman's comment above leads me to believe that it does. I am trying to determine the feasibility of migrating some of our code to modules, though, so the incompatibility concerns me.

dilipkrish:
[...]

PR #2869, opened 2019-01-16
[WIP][#2064] Relocate split-package classes to distinct packages.

Syquel:
As a prerequisite to modularize Springfox and make it therefore Java 9+ compatible, split-packages must be removed.
The approach is detailed out in: https://github.com/springfox/springfox/issues/2064#issuecomment-448181692


jithintkuriakose:
Is there any update on this issue. Using https://jitpack.io/#valery1707/springfox-assembly/2.9.2 doesn't solved my problem...

Syquel:
The maintainer never responded to my comment in that ticket and I won't work further on this until he gives his go for my approach.

dilipkrish:
@Syquel Would you like to get on a Skype sometime to figure this out. I really think this is much needed to be done and thank you for helping out with this.

dilipkrish:
@dschulten some of this impacts the integration support you add. Im curious if you could take a cursory look at this and let us know if you have any feedback. Any input would be much appreciated.

PR #2874, opened 2019-01-19
Inherit implicit params from class

### LINKED ISSUE #2550: Add support for @ApiImplicitParams at class level
- [x] What version of the library are you using? 2.9.2

What kind of issue is this?
 - [x] Feature Request

As already reported in #1659 and remarked in its latest comments, we need support for `@ApiImplicitParams` at class level, although we already have `Docket#globalOperationParameter()`.

In fact, while the latter lets add parameters to every endpoint, supporting that annotation would allow declaring parameters just for all the endpoints within a _single_ controller.

andrzejressel:
Closes #2550


khayyamsaleem:
Is there any movement on this PR? What would it take to proceed here?

PR #2897, opened 2019-02-08
Fix #2622 Error on use X-Forwarded-Prefix and ForwardedHeaderFilter

### LINKED ISSUE #2622: Error on use X-Forwarded-Prefix and ForwardedHeaderFilter
Hi

When use X-Forwarded-Prefix (Over a zuul/cloud-gateway proxy) and ForwardedHeaderFilter, what remove the header an put on context path of ServletRequest the json the base path is not updated, the path provider extension is use only on startup, not runtime 

Can I do same workaround to solve this problem? or send a PR with a new feature?

rainoko:


rainoko:
Is there hope for merge?

vkennke:
@dilipkrish any news on that?

Issue #2761, opened 2018-11-05
Pageable support should not be only springfox-data-rest


cn-src:
If I don't use `springfox-data-rest`, I won't be able to get support for `org.springframework.data.domain.Pageable`.

For example, I only use `spring-data-jpa`.

PR #2943, opened 2019-03-17
Spring pageable support

Zomzog:
#### What's this PR do/fix?

Add spring `Pageable` configuration support for #2761

The unit test is partial

I've added a hack for adding description and example. I think it's useful but I can remove it.

WIthout this config, generated swagger contain invalid fields and generate noncompilable code with [openapi-generator](https://github.com/OpenAPITools/openapi-generator)

### DIFF
--- a/springfox-spring-integration/build.gradle
+++ b/springfox-spring-integration/build.gradle
@@ -17,11 +17,13 @@ dependencies {
   api project(':springfox-core')
   api project(':springfox-spring-web')
   api project(':springfox-schema')
+  api project(':springfox-swagger-common')
 
   implementation "com.athaydes.rawhttp:rawhttp-core:2.2.1"
   compileOnly libs.springProvidedWithSpringIntegration
   compileOnly "org.springframework.restdocs:spring-restdocs-mockmvc:2.0.4.RELEASE"
   compileOnly libs.clientProvided
+  compileOnly "org.springframework.data:spring-data-rest-webmvc:$springDataRest"
 
   testImplementation libs.test
   testImplementation libs.swagger2Core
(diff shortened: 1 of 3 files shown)


springfox-spring-integration/build.gradle
```diff
@@ -16,8 +16,10 @@ dependencies {
   compile project(':springfox-spi')
   compile project(':springfox-schema').sourceSets.main.output
   compile project(':springfox-spring-web')
+  compile project(':springfox-swagger-common')
```
dilipkrish: Why do we need this dependency?

springfox-spring-integration/build.gradle
```diff
@@ -16,8 +16,10 @@ dependencies {
...
   compile "com.athaydes.rawhttp:rawhttp-core:2.0"
   compileOnly "org.springframework.restdocs:spring-restdocs-mockmvc:2.0.3.RELEASE"
+  compileOnly "org.springframework.data:spring-data-rest-webmvc:$springDataRest"
```
dilipkrish, line 26: Can this be provided?

Zomzog: Not sure to understand what you mean by "provided".

springfox-spring-integration/build.gradle
Zomzog: It's for adding annotation but maybe I can remove them.

Zomzog:
@dilipkrish I've rebased it, I will try to answer faster this time.

ryandanielspmc:
@Zomzog @dilipkrish One issue I see here is that this code uses the word `limit`, whereas, according to this article: https://www.baeldung.com/spring-data-web-support and this article: https://reflectoring.io/spring-boot-paging/, the parameter created by spring is called `size`.

Zomzog:
> @Zomzog @dilipkrish One issue I see here is that this code uses the word `limit`, whereas, according to this article: https://www.baeldung.com/spring-data-web-support and this article: https://reflectoring.io/spring-boot-paging/, the parameter created by spring is called `size`.

good point, I've fixed the name.

Issue #2763, opened 2018-11-06
Support spring-context-indexer or Provide auto configuration In Spring 5


wreulicke:
- [x] What version of the library are you using? Is it the latest version? The latest released version is [ ![Download](https://api.bintray.com/packages/springfox/maven-repo/springfox/images/download.svg) ](https://bintray.com/springfox/maven-repo/springfox/_latestVersion)

This project uses ComponentScan like [here: Swagger2DocumentationWebMvcConfiguration](https://github.com/springfox/springfox/blob/master/springfox-swagger2/src/main/java/springfox/documentation/swagger2/configuration/Swagger2DocumentationWebMvcConfiguration.java#L42)
But, if we use [spring-context-indexer](https://docs.spring.io/spring/docs/current/spring-framework-reference/core.html#beans-scanning-index), we can not find bean.

So, Will we provides `spring.components` using spring-contex-indexer, or provides auto configuration class.

In my opinion, it is happy to provide auto configuration
Even if you **do not** use spring-contex-indexer, you can not find bean when provides `spring.components` using spring-contex-indexer.

dilipkrish:
@wreulicke Im not familiar with context-indexer. Would love for you to describe the solution, or better still would appreciate a PR.

atsu85:
@wreulicke, either i don't unederstand following statement
> Even if you do not use spring-contex-indexer, you can not find bean when provides spring.components using spring-contex-indexer.

exactly, or this is wrong, because if i remove 
`annotationProcessor 'org.springframework:spring-context-indexer'`
from my project dependencies (that uses springfox-swagger), then application starts up without issues described in #2451:
`Parameter 2 of method swagger2ControllerMapping in springfox.documentation.swagger2.configuration.Swagger2DocumentationConfiguration required a bean of type 'springfox.documentation.swagger2.mappers.ServiceModelToSwagger2Mapper' that could not be found.`

atsu85:
> if i remove
`annotationProcessor 'org.springframework:spring-context-indexer'`
from my project dependencies (that uses springfox-swagger), then application starts up without issues

Note that is true if there are no jars with spring.components file in classpath. Othwrise spring 5 disables classpath scanning (unless indexer is expclitily disabled)

atsu85:
[...]

PR #2971, opened 2019-04-13
Remove ComponentScan from springfox-swagger2

### LINKED ISSUE #2451: Springfox swagger2 component does not get indexed by Spring Framework's 5 `component-indexer`
# **Springfox Swagger Version**

2.8.0

# **Spring Boot Version**

2.0.2.RELEASE

# **Description**
When using `component-indexer` with `springfox-swagger2`, the application fails to start.

Error:
```
Parameter 2 of method swagger2ControllerMapping in springfox.documentation.swagger2.configuration.Swagger2DocumentationConfiguration 
required a bean of type 'springfox.documentation.swagger2.mappers.ServiceModelToSwagger2Mapper' 
that could not be found.
```

# **Test case**

https://github.com/ssouris/swagger-component-indexer-bug

wreulicke:
Remove ComponentScan from swagger2 configuration class.

#### What's this PR do/fix?

Fixes #2451 and #2763.

#2451 and #2763.


wreulicke:
@dilipkrish Cloud you review this PR?

lucasoares:
I think you should transform ComponentScan of `SpringfoxWebMvcConfiguration` and 
`SwaggerCommonConfiguration` to auto configuration classes as well, right?

wreulicke:
> I think you should transform ComponentScan of `SpringfoxWebMvcConfiguration` and
> `SwaggerCommonConfiguration` to auto configuration classes as well, right?

I would like to remove ComponentScan and i would like to use like `spring-swagger2-autoconfigure`.
But, this PR should not make the configuration class to register auto configuration.
Because that changes is big changes and the changes effects widely.
so, I would like to separate issue for `ComponentScan` and `auto-configuration`.

lucasoares:
> > I think you should transform ComponentScan of `SpringfoxWebMvcConfiguration` and
> > `SwaggerCommonConfiguration` to auto configuration classes as well, right?
> 
> I would like to remove ComponentScan and i would like to use like `spring-swagger2-autoconfigure`.
> But, this PR should not make the configuration class to register auto configuration.
> Because that changes is big changes and the changes effects widely.
> so, I would like to separate issue for `ComponentScan` and `auto-configuration`.

You are correct.. But I'm afraid this PR will not necessarily solve the problem itself. But it is a great start.

wreulicke:
@lucasoares 
Could you try this branch with https://jitpack.io/#wreulicke/springfox?

lucasoares:
@wreulicke hello, I'm really deep into a project right now and I don't know when I will have time to try it..

I will try it as soon as I can.

build.gradle
```diff
@@ -16,6 +16,7 @@ buildscript {
     classpath "org.asciidoctor:asciidoctor-gradle-plugin:1.6.0"
     classpath 'org.ajoberstar.grgit:grgit-core:3.0.0'
     classpath 'org.ajoberstar:gradle-git-publish:2.0.0'
+    classpath "gradle.plugin.org.jetbrains.gradle.plugin.idea-ext:gradle-idea-ext:0.5"
```
dilipkrish: Doesnt using `https://plugins.gradle.org/plugin/net.ltgt.apt-idea` fix this for you in the IDE automatically?

atsu85:
@wreulicke, as i understand it would be much more easier to resolve the same issue (#2763 and #2451) by just generating `META-INF/spring.components` using
```gradle
dependencies {
    annotationProcessor 'org.springframework:spring-context-indexer'
    testAnnotationProcessor 'org.springframework:spring-context-indexer'
}
```
for all springfox related jars instead of all changes in this PR. [See related spring doc](https://docs.spring.io/spring/docs/5.2.x/spring-framework-reference/core.html#beans-scanning-index)

## Benefits:
* this library could still remain using `@Component` instead of defining beans manually using `@Bean`
  * less boilerplate
  * no need to change coding rules
* application startup time may be shorter for Spring 5, as it would use generated `spring.components` instead of classpath scanning (even if the project itself doesn't use `spring-context-indexer`)
* projects that want to use this library could also use `spring-context-indexer`
    * without getting errors related to missing beans for classes that are annotated with `@Component` (such as `Parameter 2 of method swagger2ControllerMapping in springfox.documentation.swagger2.configuration.Swagger2DocumentationConfiguration required a bean of type 'springfox.documentation.swagger2.mappers.ServiceModelToSwagger2Mapper' that could not be found.`)
* for applications that use Spring versions older than 5 everything would work the same way (jar file just contains additional `META-INF/spring.components` that is ignored for older versions)

build.gradle
wreulicke: I seem that `net.ltgt.apt-idea` is no longer maintained.

ref: https://github.com/tbroyer/gradle-apt-plugin/blob/master/README.md

wreulicke: And we need this plugin [here](https://github.com/springfox/springfox/pull/2971/files#diff-ed433482e37889a2349d935928977bb3R9-R15)

springfox-swagger2/build.gradle
```diff
@@ -39,10 +39,19 @@ dependencies {
   testCompile libs.test
 }
 
+idea {
+
+  module {
+    sourceDirs += project.file('out/production/classes/generated')
+    generatedSourceDirs += project.file('out/production/classes/generated')
+    testSourceDirs += project.file('out/production/classes/generated_tests')
+  }
```
wreulicke: We require here.
These settings enable to import classes generated by MapStruct.

build.gradle
wreulicke: @dilipkrish Could you review again?

PR #3050, opened 2019-07-16
Feature/fix 3021 required param ignored

### LINKED ISSUE #3021: ApiParamParameterBuilder ignores ParameterRequiredReader
Hello.

I'm using v2.8.0 but I had tested v2.9.2.

My API looks like below:

```java
@GetMapping("/api/users/{userNo}")
public User getUser(@PathVariable long userNo) {
    ...
}
```

And I tried to add @ApiParam

```java
@GetMapping("/api/users/{userNo}")
public User getUser(@ApiParam("${api.users.getUser.params.userNo}") @PathVariable long userNo) {
    ...
}
```

If I add the @ApiParam annotation, the userNo parameter is not marked as required in Swagger UI.

The order of the parameterPlugins that DocumentationPluginsManager configured is:
```
plugins = {$Proxy151@10640}  size = 7
 0 = {ParameterMultiplesReader@10642} 
 1 = {ParameterTypeReader@10643} 
 2 = {ParameterRequiredReader@10644} 
 3 = {ParameterDataTypeReader@10645} 
 4 = {ParameterDefaultReader@10646} 
 5 = {ParameterNameReader@10647} 
 6 = {ApiParamParameterBuilder@10648} 
```
According to above the orders, apply() of ParameterRequiredReader is invoked earlier than ApiParamParameterBuilder. So ParameterRequiredReader sets the required of the parameter builder to true.
However ApiParamParameterBuilder sets the property again to false.

I want to do not use the required property of @ApiParam because there is @PathVariable already.

How can I do?

Thanks.

kfiles:
Fixes #3021 
The order of the plugins being applied meant that the ApiParam plugin (which defaults required=false) always took precedence over the ParameterRequiredReader plugin.
I also noticed that the NotNull annotation was not supported, despite the docs indicating it should be used to indicate required parameters.


springfox-spring-web/src/main/java/springfox/documentation/spring/web/readers/parameter/ParameterRequiredReader.java
```diff
@@ -42,10 +43,12 @@
 import java.util.Set;
 import java.util.function.Predicate;
 
+import javax.validation.constraints.NotNull;
+
 import static java.util.Optional.*;
 
 @Component
-@Order(Ordered.HIGHEST_PRECEDENCE)
+@Order(Ordered.LOWEST_PRECEDENCE)
```
kfiles: In order for this plugin to control whether the parameter is required, it must be applied last, not first.

springfox-spring-web/src/main/java/springfox/documentation/spring/web/readers/parameter/ParameterRequiredReader.java
```diff
@@ -118,7 +127,8 @@ private boolean optionalButPresentInThePath(
 
   @SuppressWarnings("squid:S1872")
   boolean isOptional(ResolvedMethodParameter methodParameter) {
-    return "com.google.common.base.Optional".equals(methodParameter.getParameterType().getErasedType().getName());
+    String paramType = methodParameter.getParameterType().getErasedType().getName();
+    return "com.google.common.base.Optional".equals(paramType) || "java.util.Optional".equals(paramType);
```
kfiles: We need to support both Guava and JSR Optional, now

swagger-contract-tests/src/test/resources/contract/swagger/declaration-pet-service.json
```diff
@@ -304,7 +304,7 @@
                             "paramType": "query",
                             "format": "int32",
                             "type": "integer",
-                            "required": false
+                            "required": true
```
kfiles: A number of these tests had invalid contract specs, as a result of the prior incorrect behavior of required parameters.

springfox-spring-web/build.gradle
```diff
@@ -13,6 +13,8 @@ dependencies {
   compile "io.github.classgraph:classgraph:$classGraph"
   compile libs.core
   compile libs.spring
+  compile libs.swagger2Core
```
dilipkrish: This is not the correct fix. There is a reason that swagger is not part of this spring-web library. We don't want to mix that concern in this library. For e.g. what about open api annotations. In that case this will become a kitchen sink of stuff and loses the efficacy of having plugins.

springfox-spring-web/build.gradle
```diff
@@ -13,6 +13,8 @@ dependencies {
...
+  compile "javax.validation:validation-api:1.1.0.Final"
```
dilipkrish: Same with this. This belongs in the bean validation library.

springfox-spring-web/src/main/java/springfox/documentation/spring/web/readers/parameter/ParameterRequiredReader.java
dilipkrish: This is not desirable. The spring web has to be the highest precedence since it should be applied first.

springfox-spring-web/src/main/java/springfox/documentation/spring/web/readers/parameter/ParameterRequiredReader.java
```diff
@@ -99,6 +105,9 @@ private boolean isRequired(
       }
     }
 
+    Optional<ApiParam> apiParam = methodParameter.findAnnotation(ApiParam.class);
+    apiParam.ifPresent(param -> requiredSet.add(param.required()));
+
```
dilipkrish: This should be in its own plugin

springfox-spring-web/src/main/java/springfox/documentation/spring/web/readers/parameter/ParameterRequiredReader.java
```diff
@@ -88,6 +91,9 @@ private boolean isRequired(
     Optional<RequestHeader> requestHeader = methodParameter.findAnnotation(RequestHeader.class);
     requestHeader.ifPresent(header -> requiredSet.add(!optional && header.required()));
 
+    Optional<NotNull> notNull = methodParameter.findAnnotation(NotNull.class);
+    notNull.ifPresent(nonnull -> requiredSet.add(true));
+
```
dilipkrish: This should be in the bean validation plugin

kfiles:
> Thank you for this fix. However this isn't the correct fix.

Thanks for the response to my PR. I am happy to incorporate changes. I am definitely new to this codebase. The first question is how to fix the underlying problem. If each plugin modifies the ParameterContext, with the ParameterRequiredReader applied first and the ApiParam plugin applied last, with defaults which override any preceding plugin, how can we achieve the required semantics to satisfy PathParams? It appears that the current plugin mechanics only work if all plugins are truly orthogonal, operating on different properties of the API document model. However, a property like "required" is affected by several plugins, with none able to determine which should take precedence.

dilipkrish:
@kfiles  the way I see it is they get applied in this order
Spring MVC annotations get first precedence
Then the JSR 303 (nullable, size etc.) annotations
Then the swagger annotations

They are in the order of most generic to most specific. With each layer of plugin you get more confidence  in the values they contribute as they are more specific. The hard part is sometimes the annotations dont always have the right defaults. So what you're asking is for those defaults to be ignored in certain cases. Otherwise we're going to see the code littered with more annotation and thats exactly what we dont want philosophically in this library. Its defenitely worth looking at how to make that better. Open to ideas.

Issue #2881, opened 2019-01-25
Slow scanning time on SpringMVC application


robertoamadeuneto:
Hey guys!

I'm currently using the 2.9.2 version on Springfox and I'm facing long wait times deploying my application, just when springfox starts to scan for api listing references and generating unique operations. If I add "nickname" on @ApiOperation to avoid the unique operation generation, the slow time stays the same. 

This is my SwaggerConfiguration.java:

```
@Configuration
@EnableSwagger2
public class SwaggerConfiguration extends WebMvcConfigurerAdapter {

    private static final String RESOURCE_PACKAGE = "br.com.finchsolucoes.xgracco.application.api.resource";

    @Override
    public void addResourceHandlers(ResourceHandlerRegistry registry) {
        registry.addResourceHandler("swagger-ui.html").addResourceLocations("classpath:/META-INF/resources/");
        registry.addResourceHandler("/webjars/**").addResourceLocations("classpath:/META-INF/resources/webjars/");
    }

    @Bean
    private Docket api() {
            return new Docket(DocumentationType.SWAGGER_2)
                    .useDefaultResponseMessages(false)
                    .groupName("Api")
                    .select()
                    .apis(RequestHandlerSelectors.basePackage(RESOURCE_PACKAGE))
                    .apis(RequestHandlerSelectors.withClassAnnotation(Api.class))
                    .apis(RequestHandlerSelectors.withMethodAnnotation(ApiOperation.class))
                    .paths(regex( "/api.*"))
                    .build()
                    .ignoredParameterTypes(AuthenticationPrincipal.class)
                    .globalOperationParameters(
                        Collections.singletonList(
                                new ParameterBuilder()
                                        .name("Authorization")
                                        .description("Bearer token")
                                        .modelRef(new ModelRef("string"))
                                        .parameterType("header")
                                        .required(true)
                                        .build()));
                    .apiInfo(metaDados());
        }

        private ApiInfo metaDados() {
            return new ApiInfoBuilder()
                    .title("X-Gracco API")
                    .description("Módulo Gestão Processos.")
                    .version("5.4.0")
                    .contact(new Contact("Finch Soluções", "http://www.finchsolucoes.com.br",     "comercial@finchsolucoes.com.br"))
                    .build();
    }
}
```

[...]

PR #3098, opened 2019-08-26
Issue #2881. Do not resolve dependencies which have already been reso…

sgri:
Improves startup time by 100 times in verson 2.9.2. The problem was that dependencies were resolved multiple times for the same models.
The pull request does not target version 3.0.0, it is based on tag 2.9.2 and creates a new version 2.9.3-SNAPSHOT which is assumed to be released as 2.9.3.
Since there is no branch for 2.9.3 in this repository, I have indicated "master" as target but that should be a branch for 2.9.3 based on 2.9.2.

Version 3.0.0-SNAPHSOT creates another even worse bottleneck in springfox.documentation.spring.web.scanners.ApiModelReader#mergeModelBranch which 
requres a separate fix.


springfox-schema/src/main/java/springfox/documentation/schema/DefaultModelDependencyProvider.java
```diff
@@ -35,11 +35,8 @@
 import springfox.documentation.spi.schema.EnumTypeDeterminer;
 import springfox.documentation.spi.schema.contexts.ModelContext;
 
-import java.util.ArrayList;
-import java.util.Collection;
-import java.util.List;
-import java.util.Map;
-import java.util.Set;
+import java.util.*;
```
hannes-angst: Please use explicit imports instead of `*` imports.

sgri: > Please use explicit imports instead of `*` imports.

Done, fixed imports.

Please do not forget that the target of this pull request is not master branch, I aim to improve 2.9.2. So please create a new branch based on 2.9.2, name it 2.9.3 and then merge this pull request  into 2.9.3, then release 2.9.3.

aakhmerov:
Good day, we have noticed that this would be really useful for our project and would like to assist with review\merge process. Is there any guideline on what is left to do? would it help if we resolve conflicts and create a new PR based on this one?

roedde:
Hello,
we've the same problem in one of our projects, so it would be great if the pullrequest could be merged. The problem gets larger the more subclasses of an abstract class exists. Since we have a tranferApi which uses some meta attributes and the content is a generic parameter we have > 10 subclasses of the Abstract DTO and the application currently doesn't start within exeptable time. short example of the recursive structure:

```
@ApiModel(subTypes = {
        ConcreteDTO1.class,
        ConcreteDTO2.class,
        ConcreteDTO3.class,
}, discriminator = "disc")
public abstract class AbstractDTO<T> {

    String disc;
    int version;
    T content;
    @JsonProperty("objects")
    @ApiModelProperty(dataType = "List[AbstractDTO]")
    private List<AbstractDTO<?>> objects;
}
```

agarg-ontic (approved):

sgri:
@agarg-ontic, could you create a branch in Git based on tag  2.9.2 ? Then I will edit this pull request and set the new branch as destination. This patch addresses long initialization in version 2.9.2 and creates a new Maven version 2.9.3.
Since then codebase has changed and addressing this issue in version 3.0.0 requires a separate fix.

Issue #182, opened 2013-12-04
Duplicate class names in different packages get squashed in ControllerDocumentation - modelMap


mbazos:
I haven't yet made a fix for this yet but our team encountered this issue and I believe it's fairly easy to fix but wanted to get feedback from you on how to go about this.  

Example of the issue given the following controller method:

```
    public void sampleMethod3(com.mangofactory.swagger.spring.sample.duplicate.Pet petOne, com.mangofactory.swagger.spring.sample.Pet petTwo) {
    }
```

Notice that petOne and petTwo are different classes but have the same class name.  When the ControllerDocumnetation modelMap is being populated it uses a friendly class name as the key so it would be "Pet" in this instance.  With "Pet" being the same key in the model map they will overwrite each other.

Ways I was thinking to go about fixing this:
1.  Just add in the package name...so friendly names will now be fully qualified (Not so friendly names)
2.  Add configuration into SwaggerConfiguration that would trigger friendly/fully qualified names

I am really not sure which one to do...friendly names are nice but people might end up not realizing this is happening so it's almost safer to have fully qualified.

Not sure if you need a reason of why this is important but here is it:
1.  Service versioning - having multiple version of client model object could cause these collision to happen over time
2.  Services that return subsets of a canonical model - some data might be expensive or unnecessary to return in this case service A could return petOne (superset) and service B could return petTwo (subset of petOne).

Let me know what you think and if I have time I don't mind working on this fix as it will benefit our team.

donovanmuller:
I have a similar situation except that this is happening on the response class as well.
We have some generated JAXB classes which unfortunately share the same top level class. Even though the classes are fully qualified, springfox seems to pick one and use that everywhere.

Example below:

```
@ApiOperation(value = "Save something")
public blah.blah.Document create(@RequestBody foo.bar.Document thingy) {
  ...
  return new foo.bar.Document();
}
```

The Swagger docs will only contain one representation of `Document`...

dilipkrish:
Yeah this needs to be fixed.

stephenhibbert:
I'm also having the same issue. A fix would be very much appreciated!

mbazos:
If I have some time I might work on this.  I recently upgraded our apps from swagger-springmvc to springfox and I am really liking the re-organization and re-packaging of the libraries.

I am going to be out of the country and won't be back until the last week in September, I will make sure to check here before I start any work on this.

dilipkrish:
[...]

PR #3132, opened 2019-10-05
Fix method sort order (#182 )

MaksimOrlov:
If controller class contains overloaded methods, or methods types with same names it leads to instability of a sort order and influence to output definitions, it changes from run to run.

Spring Data Rest bean names for AssociationAction in some cases can not be defined. In that case Entity class name will be used.

That PR fixes all cases that could lead to instability of definitions.


nbv2222:
should it be merge?

mikhilvikram:
When is this likely to be merged?

goemen:
Can this ticket be given a higher priority please

dilipkrish:
Sorry this took so long @MaksimOrlov!

Issue #3024, opened 2019-06-13
Missing Support for Bean Validation API 2.0


aburmeis:
Feature Request: Version 2.9 works well with Spring 5 but using `springfox-bean-validators` (which has a dependency on validation-api 2.0.1.Final) still supports only annotations from validation API 1.0.
SO if I annotate a string model property with `@NotEmpty` instead of `@NotNull` the expected required is missing. It would be great if new annotations would also be taken into considerations generating the definition types:

* `@NotEmpty`and `@NotBlank` for required
* `@Positive`, `@PositiveOrZero`, `@Negative`, `@NegativeOrZero` for minimum/maximum/... as Size, Min and Max
* `@Email` for format

MaksimOrlov:
@aburmeis , I'm going to add all missing annotations as a part of work for jsr-303 validation groups support. Will try add that to the next release.

himanshugarg574:
Hi @MaksimOrlov , any ETA on this? Thanks

MaksimOrlov:
@himanshugarg574, there is a lot of [work]( https://github.com/springfox/springfox/tree/feature/2022/oas-3-support-experiment1) have already done. We are trying to force release version.

PR #3133, opened 2019-10-06
[feature/bean-validation-positive-negative]: Adds compatibility for t…

sanmibuh:
…hose new bean validation 2.0.1 annotatios:

* @Positive
* @PositiveOrZero
* @Negative
* @NegativeOrZero

This PR is a part of the issue #3024

### DIFF
--- a/springfox-bean-validators/src/main/java/springfox/bean/validators/plugins/schema/MinMaxAnnotationPlugin.java
+++ b/springfox-bean-validators/src/main/java/springfox/bean/validators/plugins/schema/MinMaxAnnotationPlugin.java
@@ -19,19 +19,23 @@
 package springfox.bean.validators.plugins.schema;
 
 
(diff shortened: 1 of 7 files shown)


springfox-bean-validators/src/main/java/springfox/bean/validators/plugins/schema/MinMaxAnnotationPlugin.java
```diff
@@ -19,20 +19,24 @@
 package springfox.bean.validators.plugins.schema;
 
 
+import static springfox.bean.validators.plugins.RangeAnnotations.allowableRange;
+import static springfox.bean.validators.plugins.Validators.annotationFromBean;
+import static springfox.bean.validators.plugins.Validators.annotationFromField;
```
dilipkrish: Imports need to be splatted per existing conventions

Issue #3157, opened 2019-10-23
x-forwarded-prefix ignored with springfox webflux


jgcollings:
**Using springfox 3.0.0-SNAPSHOT in a Spring Webflux project
When providing X-Forwarded headers in request to /v2/api-docs, the spec that is generated always has a basePath of "/"**

### Steps to reproduce:

1 - clone repo https://github.com/jgcollings/webfluxtest.git
2 - run the project: mvn spring-boot:run
3 - use the following curl command to get the api spec:
```
curl -X GET \
  http://localhost:8080/v2/api-docs \
  -H 'X-Forwarded-Host: www.acme.com' \
  -H 'X-Forwarded-Port: 80' \
  -H 'X-Forwarded-Prefix: /myprefix' \
  -H 'X-Forwarded-Protocol: http'
```

Notice that the basePath is "/" instead of "/myprefix"

The problem seems to be in Swagger2ControllerWebFlux.java line 91
`swagger.basePath("/");`
I believe the issue would be solved with this code change:
`swagger.basePath(isEmpty(request.getPath().contextPath().value()) ? "/" : request.getPath().contextPath().value());`

jgcollings:
Could I get someone to acknowledge this issue and merge the PR with the proposed fix?

PR #3158, opened 2019-10-23
Set the basePath using the request context path if present

jgcollings:
This allows the X-Forwarded-Prefix header to work correctly

#### fix for issue #3157

Test with the following steps:
1 - build and install into local maven repo.
2 - clone repo https://github.com/jgcollings/webfluxtest.git
3 - run the project: mvn spring-boot:run
4 - use the following curl command to get the api spec:

curl -X GET \
  http://localhost:8080/v2/api-docs \
  -H 'X-Forwarded-Host: www.acme.com' \
  -H 'X-Forwarded-Port: 80' \
  -H 'X-Forwarded-Prefix: /myprefix' \
  -H 'X-Forwarded-Protocol: http'

Notice that the basePath in the spec is "/myprefix"


M3yo:
Hi everyone,

When this fix will be merged ?

Both controllers (webMvc and WebFlux) does not work the same regarding basePath, the X-Forwarded-Prefix is ignored in Webflux mode which make the "Try it out" feature unusable when using a reverse proxy. :(

jgcollings:
I would like this merged soon as well. It has been a problem for us for quite a while and I have to tell developers constantly that the try it out button does not work.

PR #3175, opened 2019-11-06
Create a FormParameter model object instead of BodyParameter

kremers:
Based on Sebastian Hartte / shartte #2722
Resolved conflicts with current master


kremers:
It seems reopening does not trigger a new build :(

Issue #3189, opened 2019-11-19
Spring property placeholder - default values not handled


beliakou:
Using 2.9.2:

If `@RequestParam.defaultValue` contains Spring placeholder with default value, e.g. `${propertyName:defaultValue}` it is not handled properly:

```
java.lang.NumberFormatException: For input string: "${propertyName:100}"
        at java.lang.NumberFormatException.forInputString(NumberFormatException.java:65) ~[?:?]
        at java.lang.Long.parseLong(Long.java:678) ~[?:?]
        at java.lang.Long.valueOf(Long.java:1144) ~[?:?]
        at io.swagger.models.parameters.AbstractSerializableParameter.getDefault(AbstractSerializableParameter.java:322) [swagger-models-1.5.20.jar:1.5.20]
        at jdk.internal.reflect.GeneratedMethodAccessor112.invoke(Unknown Source) ~[?:?]
        at jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43) ~[?:?]
        at java.lang.reflect.Method.invoke(Method.java:566) ~[?:?]
        at com.fasterxml.jackson.databind.ser.BeanPropertyWriter.serializeAsField(BeanPropertyWriter.java:688) [jackson-databind-2.9.9.jar:2.9.9]
        at com.fasterxml.jackson.databind.ser.std.BeanSerializerBase.serializeFields(BeanSerializerBase.java:719) [jackson-databind-2.9.9.jar:2.9.9]
        at com.fasterxml.jackson.databind.ser.BeanSerializer.serialize(BeanSerializer.java:155) [jackson-databind-2.9.9.jar:2.9.9]
        at com.fasterxml.jackson.databind.ser.impl.IndexedListSerializer.serializeContents(IndexedListSerializer.java:119) [jackson-databind-2.9.9.jar:2.9.9]
        at com.fasterxml.jackson.databind.ser.impl.IndexedListSerializer.serialize(IndexedListSerializer.java:79) [jackson-databind-2.9.9.jar:2.9.9]
        at com.fasterxml.jackson.databind.ser.impl.IndexedListSerializer.serialize(IndexedListSerializer.java:18) [jackson-databind-2.9.9.jar:2.9.9]
        at com.fasterxml.jackson.databind.ser.BeanPropertyWriter.serializeAsField(BeanPropertyWriter.java:727) [jackson-databind-2.9.9.jar:2.9.9]
[...]

PR #3190, opened 2019-11-19
#3189 - handle defaults in Spring placeholders

beliakou:
#### What's this PR do/fix?
Issue #3189

Few test cases added

As per http://springfox.github.io/springfox/docs/snapshot/#property-file-lookup there is support for Spring properties resolution but default values like `${propertyName:defualt}` are not handled. I'm wondering if there are other places I need to update?

Issue #3189


dilipkrish:
Thank you @beliakou!

Issue #3255, opened 2020-02-18
Springfox Bean Validations @Size annotation does not work properly on List


shouldnotappearcalm:
Please take the time to search the repository, if your question has already been asked or answered.

- [ ] What version of the library are you using? `2.9.2` 

## Question

I am using springfox bean validators to extract field limit information into the generated swagger json file，Here is the way I use:

```Java
    @ApiModelProperty("str1")
    @Size(min = 1, max = 200, message = "length 1-200")
    private String str1;
    
    @ApiModelProperty("list1")
    @Size(min = 1, max = 100, message = "size 1-100")
    private List<String> list1;
```
In the generated json file：

```json
str1:- {
  type: "string",
  description: "str1",
  allowEmptyValue: false,
  minLength: 1,
  maxLength: 200
},
list1:- {
  type: "array",
  description: "list1",
  allowEmptyValue: false,
  items:- {
    type: "string"
  }
}
```
`@Size` annotation does not work properly on `List`

PR #3259, opened 2020-02-23
Issue 3255

shouldnotappearcalm:
#### What's this PR do/fix?
fix `@Size` annotation does not work properly on `List`
You can write the following code in the bean
```Java
@ApiModelProperty("list1")
@Size(min = 1, max = 200, message = "1-200")
private List<String> list1;
```
Then you can see in the generated swagger json
```json
list1:- {
type: "array",
description: "list1",
items:- {
  type: "string"
},
maxItems: 200,
minItems: 1
}
```

As above
[3255](https://github.com/springfox/springfox/issues/3255)

### DIFF
--- a/springfox-swagger2/src/main/java/springfox/documentation/swagger2/mappers/EnumMapper.java
+++ b/springfox-swagger2/src/main/java/springfox/documentation/swagger2/mappers/EnumMapper.java
@@ -21,7 +21,14 @@ package springfox.documentation.swagger2.mappers;
 
 import io.swagger.models.ModelImpl;
 import io.swagger.models.parameters.SerializableParameter;
-import io.swagger.models.properties.*;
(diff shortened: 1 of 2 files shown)


shouldnotappearcalm:
I looked at the test methods in the `EnumMapperSpec` class, but I feel that `ArrayProperty` class does not seem to be particularly suitable for current methods, because it does not have properties such as `enums `and `minimum`. Do I need to add a test method? @dilipkrish

dilipkrish:
You can add a new method for your test. Without that there is no safety net for your change.

shouldnotappearcalm:
Hi,I have added the corresponding unit tests for the new code. Please take a look when you are free. @dilipkrish

dilipkrish:
Thank you @shouldnotappearcalm

Issue #3301, opened 2020-05-05
Schema for class that is a subtype of multiple classes only take into account one of the parent interface


sugiteow-encompass:
What version of the library are you using? Is it the latest version?
2.9.2

What kind of issue is this?
Bug Report

**Steps to Reproduce:** 

1.  Create a class that has multiple polymorphic parents like so: 
```
@JsonTypeInfo(use = JsonTypeInfo.Id.NAME, include = JsonTypeInfo.As.EXISTING_PROPERTY, property = "type")
@JsonSubTypes({
        @JsonSubTypes.Type(value = BothInputAndFormParameters.class, name = "BothInputAndFormParameters"),
        @JsonSubTypes.Type(value = SomeFormParameters.class, name = "SomeFormParameters")
})
public interface PolicyFormParameters {
}

@JsonTypeInfo(use = JsonTypeInfo.Id.NAME, include = JsonTypeInfo.As.EXISTING_PROPERTY, property = "type")
@JsonSubTypes({
        @JsonSubTypes.Type(value = SomeInputParameters.class, name = "SomeInputParameters")
        @JsonSubTypes.Type(value = BothInputAndFormParameters.class, name = "BothInputAndFormParameters"),
})
public interface PolicyInputParameters {
}

public class BothInputAndFormParameters implements PolicyFormParameters, PolicyInputParameters {

}

public class SomeFormParameters implements PolicyFormParameters {
}

public class SomeInputParameters implements PolicyInputParameters {
}


```
2.  Check the rendered swagger schema
3.  You'll see that only the first declared parent is taken into account (see below)
```
"BothInputAndFormParameters": {
      "title": "BothInputAndFormParameters",
      "allOf": [
        {
          "$ref": "#/definitions/PolicyFormParameters"
        },
        {
          "type": "object",
          "title": "BothInputAndFormParameters"
        }
      ]
    },
```

**Expected:**

Both parent should be taken into account, like so: 
```
"BothInputAndFormParameters": {
      "title": "BothInputAndFormParameters",
      "allOf": [
        {
          "$ref": "#/definitions/PolicyFormParameters"
        },
        {
          "$ref": "#/definitions/PolicyInputParameters"
        },
        {
          "type": "object",
          "title": "BothInputAndFormParameters"
        }
      ]
    },
```

Example repo here: https://github.com/sugiteow-encompass/springfox-multi-subtype

dilipkrish:
Will address this next release

[...]

PR #3302, opened 2020-05-05
Properly handle model for object that's declared as subtypes of multiple parent.

sugiteow-encompass:
#### What's this PR do/fix?
change InheritanceDeterminer to properly account for a model with multiple parent (instead of assuming a single model will have a single parent).

Yes

This will fix issue #3301


dilipkrish:
Investigating if this is viable to be merged

PR #3307, opened 2020-05-08
Returning cache to life.

MaksimOrlov:
This PR returns support for cache. Variable `parameterId` has been removed from `.equals()` and `.hashCode()`  methods of `ModelContext`. Speed  on load testing shows increasing from _**130**_ test controllers to _**300**_ controllers per second.

### DIFF
--- a/springfox-spring-web/src/main/java/springfox/documentation/spring/web/scanners/ApiListingScanner.java
+++ b/springfox-spring-web/src/main/java/springfox/documentation/spring/web/scanners/ApiListingScanner.java
@@ -46,7 +46,6 @@ import java.util.Map;
 import java.util.Objects;
 import java.util.Optional;
 import java.util.Set;
-import java.util.TreeMap;
 import java.util.function.Predicate;
 import java.util.stream.Stream;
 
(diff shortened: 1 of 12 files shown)


springfox-spring-web/src/main/java/springfox/documentation/spring/web/scanners/ApiListingScanner.java
```diff
@@ -127,7 +126,7 @@ public ApiListingScanner(
 
     List<SecurityReference> securityReferences = new ArrayList<>();
 
-    Map<String, Set<Model>> globalModelMap = new TreeMap<>();
+    Map<String, Set<Model>> globalModelMap = new HashMap<>();
```
dilipkrish, line 129: This was tree map because it preserves the order of the models. Does this need to change to hash map?

dilipkrish:
Thank you @MaksimOrlov!

springfox-spring-web/src/main/java/springfox/documentation/spring/web/scanners/ApiListingScanner.java
MaksimOrlov: Iterating over HashMap is faster than iterating TreeMap, and order is not important any more

Issue #3407, opened 2020-07-15
Bug: 


sigand:
implementation "io.springfox:springfox-boot-starter:3.0.0"
Spring boot version: 2.3.1.RELEASE

```
      @ApiParam(
          value = "some description",
          required = false,
          allowMultiple = true,
          collectionFormat = "csv"
      )
      val somename: List<String>?
```

Becomes 

```
{
  name: "somename",
  in: "query",
  description: "some description",
  required: false,
  type: "array",
  items: {
    type: "string"
  },
  collectionFormat: "**multi**"
}
```

It looks to be correct in ApiParamParameterBuilder, but is turned to **multi** by the time it reaches RequestParameterMapper

dilipkrish:
It should preserve the override... thanks for reporting

sigand:
I think I have a fix, not sure if it's the correct way to do it though

dilipkrish:
Its the right fix! thank you!

PR #3411, opened 2020-07-15
Bugfix/3407

sigand:
#### What's this PR do/fix?
honor collectionType as specified in 

Added endpoint to Bug controller, manually verified

This solves our specific use case, but I am not sure it is in the right place, or general enough.

https://github.com/springfox/springfox/issues/3407


dilipkrish:
Thank you for tracking that down. 🤘 Could you also please squash your commits and also update the contracts by[ commenting these lines](https://github.com/springfox/springfox/blob/a91d24e96a78c53bfcc1d534e1bd664ed1403331/swagger-contract-tests/build.gradle#L44-L47) and running the build `./gradlew :swagger-contract-tests:build` (which will update the contracts).

PR #3421, opened 2020-07-16
Update README.md

robertsearle:
I had to do this to stop it from crashing.  Still not working for me though

### DIFF
--- a/README.md
+++ b/README.md
@@ -55,6 +55,26 @@ NOTE: Would love feedback to make this better
 1. Remove explicit dependencies on `springfox-swagger2`
 2. Remove any `@EnableSwagger2...` annotations
 3. Add the `springfox-boot-starter` dependency
+I had to do this.  https://github.com/springfox/springfox/issues/2932
+```
+        <dependency>
+            <groupId>io.springfox</groupId>
+            <artifactId>springfox-boot-starter</artifactId>
+            <exclusions>
+                <exclusion>
+                    <groupId>org.springframework.plugin</groupId>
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -55,6 +55,26 @@ NOTE: Would love feedback to make this better
...
 2. Remove any `@EnableSwagger2...` annotations
 3. Add the `springfox-boot-starter` dependency
+I had to do this.  https://github.com/springfox/springfox/issues/2932
+```
+        <dependency>
+            <groupId>io.springfox</groupId>
+            <artifactId>springfox-boot-starter</artifactId>
+            <exclusions>
+                <exclusion>
+                    <groupId>org.springframework.plugin</groupId>
```
kdebski85, line 65: io.springfox::springfox-boot-starter::3.0.0 depends on org.springframework.plugin::spring-plugin-core::2.0.0.RELEASE
so you are excluding and then adding the same artifact. It does not change anything.

Which version of springfox-boot-starter do you use?

I followed 3 steps above and it worked fine for me with 3.0.0.
URL to UI changed. It is now .../swagger-ui/index.html

I think this PR should be closed without merging.

dilipkrish: I agree. There is only one version of springfox-boot-starter which is 3.0.0, but Im more interested in what @robertsearle is running into that he says is  *Still not working for me though*

dilipkrish:
@robertsearle closing this PR. Please open an issue with the problem you're facing and lets solve it together

PR #3467, opened 2020-07-30
Readme/Docs: mention @EnableWebMvc for Spring Boot migration

martin-mfg:
I upgraded from Springfox 2.9.2 to 3.0.0 in our Spring Boot application. We were using WebMvc without the `@EnableWebMvc` annotation and the swagger UI wasn't reachable. The problem was that Springfox' `SwaggerUiWebMvcConfigurer` was never called. After adding `@EnableWebMvc` to our WebMvc configuration class, http://localhost:8080/swagger-ui/index.html does now work fine.

To keep others from running into the same problem, this PR adds information about `@EnableWebMvc` to the Spring Boot migration sections in Readme and docs.

By the way, http://localhost:8080/swagger-ui/ still doesn't work in our application, but I haven't yet figured out what the problem is.


dilipkrish:
Thank you!

PR #3480, opened 2020-08-07
Fix kotlin suspend function return type

SokoMishaLov:
Since spring boot 2.2.0.RELEASE there is a possibility to use kotlin suspend functions in the webflux environment.
```kt
@GetMapping("/foo")
suspend fun foo(): String = TODO()
```

Such methods have an "erased" return type in the result JVM bytecode so it causes wrong model generation, unfortunately.
```kt
class Foo {
    suspend fun heh(): Bar = TODO()
    fun mda(): Bar = TODO()
}

class Bar

fun main() {
    Foo::class.java.declaredMethods.forEach(::println)
}

// prints
// public final java.lang.Object com.example.Foo.heh(kotlin.coroutines.Continuation)
// public final com.example.Bar com.example.Foo.mda()
```

My PR fixes it by using kotlin-reflect (optionally).


dilipkrish:
Thanks for your PR, but somethings off with this PR... I'd like to split Kotlin off into its own module and not mix it in with vanilla spring/java. While it looks logically ok, it doesn't seem to want to build :(  I'll investigate.

SokoMishaLov:
My idea was to not add kotlin code to the codebase to avoid another language to compile.
But I think you are right and I added this code to the separate module - who knows what else we should fix for kotlin in future :)

SokoMishaLov:
Well, I reveal that there is another interoperability issue with kotlin. I've force-pushed fix for detecting `kotlin.Deprecated` annotation which is actually not a typealias for `java.lang.Deprecated` so it has to handled manually.
Having a separate module was totally a good idea )

Issue #3499, opened 2020-08-16
parameter type is query,SwaggerExpandedParameterBuilder shuld not overwrite parameter name


xiejx618:
springfox version:   io.springfox:springfox-boot-starter:3.0.0,
demo code is here: https://github.com/xiejx618/demo

I'm not sure if it is a bug, but it is unreasonable.
```
BaseDto .java
import io.swagger.annotations.ApiModelProperty;
import java.io.Serializable;
public class BaseDto  implements Serializable {
    @ApiModelProperty(name = "code",value = "编码")
    private String code;
    public String getCode() {
        return code;
    }
    public BaseDto setCode(String code) {
        this.code = code;
        return this;
    }
}

UserDto.java
public class UserDto extends BaseDto{
}

TestDto.java
public class TestDto extends BaseDto{
    private UserDto userDto;
    public UserDto getUserDto() {
        return userDto;
    }
    public TestDto setUserDto(UserDto userDto) {
        this.userDto = userDto;
        return this;
    }
}

TestController.java
import org.exam.dto.TestDto;
import org.springframework.cloud.openfeign.SpringQueryMap;
import org.springframework.web.bind.annotation.GetMapping;
import org.springframework.web.bind.annotation.RestController;
@RestController
public class TestController {
    @GetMapping("test")
    public String start(@SpringQueryMap TestDto dto) {
        return "test";
    }
}
```
![image](https://user-images.githubusercontent.com/2609827/90326379-57501a80-dfba-11ea-9fb1-197020a78578.png)

When I remove the name attribute value of @ApiModelProperty, it is correct at this time
```
import io.swagger.annotations.ApiModelProperty;
import java.io.Serializable;
public class BaseDto  implements Serializable {
    @ApiModelProperty(/*name = "code",*/value = "编码")
    private String code;
    public String getCode() {
        return code;
    }
    public BaseDto setCode(String code) {
        this.code = code;
        return this;
    }
}
```
![image](https://user-images.githubusercontent.com/2609827/90326395-85355f00-dfba-11ea-8c97-f89ccefb9cb6.png)

There are two parameters here and should not be treated as one parameter. I think when the parameter type is query, the parameter name should not be overwritten here。

![image](https://user-images.githubusercontent.com/2609827/90326461-376d2680-dfbb-11ea-8f68-2a0c402cdf3c.png)

![image](https://user-images.githubusercontent.com/2609827/90326499-8b780b00-dfbb-11ea-83a3-9042d6874881.png)

PR #3490, opened 2020-08-13
null pointerexception

xiejx618:

PR #3510, opened 2020-08-20
if ParameterType is query,append ParentName to parameterName

xiejx618:
#### What's this PR do/fix?
if ParameterType is query,append ParentName to parameterName

Test code is here: https://github.com/xiejx618/demo

https://github.com/springfox/springfox/issues/3499

https://github.com/springfox/springfox/issues/3499

Because this situation is treated as the same parameter, deduplication is triggered, and then a null pointer exception is triggered
https://github.com/springfox/springfox/pull/3490


dilipkrish:
Looks like there is a check style violation

dilipkrish:
Thanks for fixing it with a PR!

PR #3523, opened 2020-09-01
[Document] Update swagger-ui URL location.

hiroshi-maekawa:
#### What's this PR do/fix?
There are some documents which refers that swagger-ui location is `http://{host}/swagger-ui.html`.
This PR fixes those issues.

no, this fixes only documents.

I wasted 1 hour to find out correct path, so I don't like to made mistakes like me.


hiroshi-maekawa:
this should be merged with #3500

dilipkrish:
Thank you for the PR!

Issue #3391, opened 2020-07-14
Model properties are alphabetically sorted (without order attribute)


dilipkrish:
looks still broken, model properties are alphabetically sorted (without order attribute). 

_Originally posted by @schrek1 in https://github.com/springfox/springfox/issues/3087#issuecomment-658113981_

santhosh-v:
order attribute also not working and it sorts alphabetically

elvisbegovic:
Might be related on yesterday's SO https://stackoverflow.com/questions/63616811/springfox-apimodelproperty-position-sorting-not-working/63628683#63628683

dilipkrish:
Plan is to address all of these in a maintenance release 3.0.1

wacaine:
> Plan is to address all of these in a maintenance release 3.0.1

When can we expect maint release?

honza-zidek:
> > Plan is to address all of these in a maintenance release 3.0.1
> 
> When can we expect maint release 3.0.1?

Any news here? :)

fuchengli1024:
Which version is working? I want to down grade.
It is a critical issue

dmtmalin:
2.9.2

fuchengli1024:
I am using 2.9.2. It has same issue

pcimring:
Any ETA on this? Thanks

liukefu2050:
I copied a separate lib to temporarily solve the problem of sorting according to the position attribute

Change your dependence
```
<dependency>

<groupId>io. springfox</groupId>

<artifactId>springfox-swagger2</artifactId>

<version>3.0.0</version>

</dependency>
```
to
```
<dependency>

<groupId>com. github. shijingsh</groupId>

<artifactId>knife4j-springfox-swagger2</artifactId>

<version>3.0.0</version>

</dependency>
```
```
public class UserVo implements java.io.Serializable {
   
    @ApiModelProperty(notes = "name",position = 1)
    private String name;
   
    @ApiModelProperty(notes = "mobile",position = 2)
    private String mobile;
    
    @ApiModelProperty(notes = "email",position = 3)
    private String email;

}
```
Code location:  https://github.com/shijingsh/xiushang-knife4j/tree/master/knife4j-springfox-swagger2

PR #3573, opened 2020-09-22
Fixed issue with model properties sorting

acmi:
#### What's this PR do/fix?
This PR fies model properties order (not alphabetically ordered now).
Yes, test case added to `oas-contract-tests`. `JSONAssert` replaced with text comparing to make sure that api returning same response, even the order of keys in objects.
No.
#3391

### DIFF
--- a/oas-contract-tests/build.gradle
+++ b/oas-contract-tests/build.gradle
@@ -28,7 +28,6 @@ dependencies {
 
 
   testImplementation libs.test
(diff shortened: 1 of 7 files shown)


FabianWilms:
@dilipkrish This change seems to fix the sorting problem for ModelProperties. Can you take a look at this PR?

yunusb:
Well the problem is indeed the TreeMap after the sorting operation on the stream. TreeMap by default use the natural order. So we lose the sorting operation by collecting in it.

Issue #3413, opened 2020-07-16
swagger-ui should respect absolute paths provided by swagger-resources


dilipkrish:
I'm seeing this bug at 3.0.0. Is this now intended functionality?

_Originally posted by @TJReinert in https://github.com/springfox/springfox/issues/2235#issuecomment-659078680_

ouaibsky:
Hi, 
I confirm I'm facing same bug.
It's blocking because we aggregate swagger into an API gateway and nothing is working anymore
Regards
Christophe

ouaibsky:
FYI: I provide a quick fix in #3596

Hccake:
@dilipkrish Could you please tell me when we can fix this problem? Our aggregate documents are not available now

steinwelberg:
Is there an update available on this issue? e.g. Is the issue actually fixed already and Is there a release date known for 3.0.1?

ajitesh-kuppa-sivakumar:
I am facing the same issue. This is because of baseUrl getting added as prefix. Please fix this asap.

Hccake:
> I am facing the same issue. This is because of baseUrl getting added as prefix. Please fix this asap.

I did a simple fix by modifying the source code and then making a substitution in the project .

You can refer to my implementation, file address：https://github.com/ballcat-projects/ballcat-samples/blob/master/ballcat-sample-admin-application/src/main/resources/META-INF/resources/webjars/springfox-swagger-ui/springfox.js

matthewarmand:
This is a pretty high-impact regression for anyone using external URLs for `SwaggerResource`s. It also seems relatively low effort to fix, especially considering it was [previously fixed](https://github.com/springfox/springfox/issues/2235#issuecomment-659081498) and just regressed in 3.0.0. The issue is also over 14 months old now.

Can someone involved with @springfox prioritize a 3.0.1 release with this fix ASAP?

ouaibsky:
it's crasy, 
the fix has been provided 1 year ago !
and not yet merged or better solution implemented !

ahmadsab95:
Any update available on this issue ?

luksbelt:
Same problem here, there is no RC to fix this yet?

ahmadsab95:
@luksbelt  - No,  try the workaround that @Hccake suggests.

DianaSensei:
I got same problem, Any update available on this issue ?

PR #3597, opened 2020-10-13
Make sure swagger is working with SwaggerRsource containing full URL

ouaibsky:
It fix issue #3413

It was working in swagger 2.x.y

It's mandatory when you aggregate many swagger resources (remote services) into one service with a unique swagger-ui (like api gateway)

Christophe

### DIFF
--- a/springfox-swagger-ui/src/web/js/springfox.js
+++ b/springfox-swagger-ui/src/web/js/springfox.js
@@ -54,7 +54,7 @@ window.onload = () => {
 
   const prependBaseUrl = (baseUrl, resources) => {
     resources.forEach(function(r) {
-      r.url = baseUrl + r.url;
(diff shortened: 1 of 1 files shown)


dilipkrish:
Thanks for your PR @ouaibsky. Have you tested this in an ALB scenario with x-forward headers set?

ouaibsky:
It has been tested with spring gateway
Can you add this into 3.0.1 please, it's clearly a regression since 2.10.5.
Thx
Chris

matthewarmand:
@dilipkrish it's been over 14 months since this regression was reported, and nearly a year now since this fix was proposed. Any update on when we can expect this fix to be released? This is of pretty high impact as it completely breaks the ability of springfox to respect external URLs.

babyhuey:
@dilipkrish @martypitt @adrianbk ?

PR #3666, opened 2020-12-05
Add support for @Validated groups

### LINKED ISSUE #2307: Conditional @Validated annotation is ignored
Please take the time to search the repository, if your question has already been asked or answered.

- [ ] What version of the library are you using? Is it the latest version? The latest released version is [ Download ](https://bintray.com/springfox/maven-repo/springfox/_latestVersion)

What kind of issue is this?
 - [ ] @Validated annotation from **org.springframework.validation.annotation** is ignored by the doc generation of swagger.
 - [ ] I'm using the **springfox-bean-validators** plugin to get **@NotNull** annotation constraints up in swagger documentation. However, those annotations from JSR-303 do work in conjunction with **@Validated** annotation from **org.springframework.validation.annotation** package to apply conditional validation constraints on the same model e.g.

In Model class (Entity.java):
```
@NotNull(message = "Field cannot be null", groups = {ConditionalValidation.class})
private String someField;
```
In Controller class:
`public void createEntity( @Validated(ConditionalValidation.class) @RequestBody Entity someEntity) {}`

- As of now, _someField_ is always showing as required in the swagger documentation no matter if the conditional group is applied or not.
- I want to see in Swagger _someField_ as required but only for the API _createEntity_ and as optional for any other API that does apply the **@Validate** without the **ConditionalValidation.class**.

meisten:
**What's this PR do/fix?**

This PR fixes #2307

1. Add support for `@Validated` groups
Supported annotations:
- `@NotNull` and `@NotNull.List`
- `@Pattern` and `@Pattern.List`
- `@DecimalMin`/`@DecimalMax` and `@DecimalMin.List`/`@DecimalMax.List`
- `@Null` and `@Null.List`
- `@Min`/`@Max` and `@Min.List`/`@Max.List`
- `@NotBlank` and `@NotBlank.List`
- `@Size` and `@Size.List`

2. `@Validated` can be applied for parameter (request model) and method (response model).

**Are there unit tests? If not how should this be manually tested?**
There are no tests at this time. I added a [demonstration project][1] with modified libraries and [the resulting schema][2] from the /v3/api.

[1]: https://github.com/meose/springfox-validation-example
[2]: https://github.com/meose/springfox-validation-example/blob/master/src/main/java/springfox/validation/example/output.json

### DIFF
--- a/springfox-bean-validators/src/main/java/springfox/bean/validators/plugins/Validators.java
+++ b/springfox-bean-validators/src/main/java/springfox/bean/validators/plugins/Validators.java
@@ -119,12 +119,11 @@ public class Validators {
     }
 
     Set<Class<?>> validatedGroupsClasses = getGroupClasses(validatedGroups.stream().map(ResolvedType::getErasedType));
-    Set<Class<?>> constraintGroupClasses = getGroupClasses(Stream.of(groupsConstraint));
+    Set<Class<?>> constraintGroupClasses = Stream.of(groupsConstraint).collect(Collectors.toSet());
 
-    // Case #2: Validated groups contains Default.class
-    // Constrain groups contains Default.class
-    if (validatedGroupsClasses.contains(Default.class)) {
-      return constraintGroupClasses.isEmpty() || constraintGroupClasses.contains(Default.class);
+    // Case #2: constraintGroupClasses is empty
+    if (constraintGroupClasses.isEmpty()) {
+      constraintGroupClasses.add(Default.class);
     }
 
     // Case #3: Validated groups does not contain Default.class
@@ -133,16 +132,18 @@ public class Validators {
   }
 
   private static Set<Class<?>> getGroupClasses(Stream<Class<?>> stream) {
-    return stream.flatMap(i -> getSuperClasses(i).stream()).collect(Collectors.toSet());
+    return stream.flatMap(i -> getAllInterfaces(i).stream()).collect(Collectors.toSet());
   }
 
-  public static Set<Class<?>> getSuperClasses(Class<?> c) {
-    Set<Class<?>> classSet = new HashSet<>();
-    classSet.add(c);
-    for (Class<?> superclass = c; superclass != null; c = superclass) {
-      classSet.add(superclass);
-      superclass = c.getSuperclass();
+  public static Set<Class<?>> getAllInterfaces(Class<?> c) {
+    Set<Class<?>> set = new HashSet<>();
+    if (!c.isInterface()) {
+      return set;
     }
-    return classSet;
+    set.add(c);
+    for (Class<?> superInterfaces : c.getInterfaces()) {
+      set.addAll(getAllInterfaces(superInterfaces));
+    }
+    return set;
   }
 }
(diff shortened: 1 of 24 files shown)


meose:
[Demonstration project][1] with modified libraries and [the resulting schema][2] from the /v3/api.

[1]: https://github.com/meose/springfox-validation-example
[2]: https://github.com/meose/springfox-validation-example/blob/master/src/main/java/springfox/validation/example/output.json

springfox-bean-validators/src/main/java/springfox/bean/validators/plugins/Validators.java
```diff
@@ -133,16 +132,18 @@ public static boolean annotationMustBeApplied(ModelPropertyContext context, Clas
   }
```
meose, line 132: ### Third condition (or `Case #3`) - 130 line

The system determines the intersection between two sets — annotation groups `@Valid` (equivalent to `@Validated(groups=Default.class)`) or `@Validated(groups=...)` and annotation groups constraints.

If an intersection exists, then the constraint annotation must be applied to this model.

springfox-bean-validators/src/main/java/springfox/bean/validators/plugins/Validators.java
```diff
@@ -119,12 +119,11 @@ public static boolean annotationMustBeApplied(ModelPropertyContext context, Clas
...
+    Set<Class<?>> constraintGroupClasses = Stream.of(groupsConstraint).collect(Collectors.toSet());
 
-    // Case #2: Validated groups contains Default.class
-    // Constrain groups contains Default.class
-    if (validatedGroupsClasses.contains(Default.class)) {
-      return constraintGroupClasses.isEmpty() || constraintGroupClasses.contains(Default.class);
+    // Case #2: constraintGroupClasses is empty
+    if (constraintGroupClasses.isEmpty()) {
+      constraintGroupClasses.add(Default.class);
     }
```
meose, line 127: ### Second condition (or `Case #2`) - 124 line
If the constraint annotation (such as `@NotNull`) does not contain a group, this case is the same as using the default group (`Default.class`).

springfox-bean-validators/src/main/java/springfox/bean/validators/plugins/Validators.java
```diff
@@ -119,12 +119,11 @@ public static boolean annotationMustBeApplied(ModelPropertyContext context, Clas
...
```
meose, line 119: ### First condition (or `Case #1`) - 114 line
If model does not have `@Valid` or `@Validated` annotation, the system should not enforce annotation constraints for this model. I think there must be some flag indicating how the system should do in this situation. Currently, the presence of annotations `@Valid` or `@Validated` does not affect the applicability of restrictions.

springfox-bean-validators/src/main/java/springfox/bean/validators/plugins/Validators.java
meose, line 127: ### Second condition (or `Case #2`) - 124 line
If the constraint annotation (such as `@NotNull`) does not contain a group, this case is the same as using the default group (`Default.class`).

springfox-bean-validators/src/main/java/springfox/bean/validators/plugins/Validators.java
meose, line 132: ### Third condition (or `Case #3`) - 130 line

The system determines the intersection between two sets — annotation groups `@Valid` (equivalent to `@Validated(groups=Default.class)`) or `@Validated(groups=...)` and annotation groups constraints.

If an intersection exists, then the constraint annotation must be applied to this model.

Issue #3502, opened 2020-08-17
Query parameters required attribute not set properly


varjasz:
Migrating from 2.9.2 to 3.0.0 and query parameters required attribute value is not set properly.

In the following Kotlin code the String? type represents a nullable parameter

```
@GetMapping
fun dummyOperation(@ApiParam("Dummy description") @RequestParam("dummyParam") dummyParam: String?) = "dummy response"

```
and in version 2.9.2 it is working properly, the swagger-ui not showing the parameter is required,
but in version 3.0.0 it shows as a required parameter (which is wrong)

I have to add @RequestParam( **required = false** ) to show it optional, but it is a redundant information, because Kotlin String? is an optional type.

Here is the decompiled Java bytecode, in the case you are not familiar with Kotlin:
```
@GetMapping
@NotNull
public String dummyOperation(@ApiParam("Dummy description") @RequestParam(value = "dummyParam") @Nullable String dummyParam) {
      return "dummy response";
   }

```

benwarfield-usds:
Probably the same issue, with a plain-java example: `@RequestParam Optional<String> filterReason` will show as a required query-string parameter, rather than as optional.

didiez:
Same problem here, using java8 `Optional`:
```
@GetMapping
public List<Event> events(@RequestParam Optional<Integer> year)
```
Is it possible to configure the param `required` flag globally based on the param type? It could work for us as a workaround in the meantime.

PR #3700, opened 2021-01-12
Fix Optional parameter detection

didiez:
#### What's this PR do/fix?
The parameter should be resolved as `required=false` when using `java.util.Optional`.
There are no unit tests.
In this commit https://github.com/springfox/springfox/commit/980379d1a042cc93594dcaf57036680924423654 there was a change from `java.util.Optional` to `com.google.common.base.Optional`. Both should be supported
#3502


didiez:
The issue is still there. This simple PR (1 line) would fix it.

PR #3720, opened 2021-02-02
fix @schema -> nullable not take effect in property

EdisonXllllll:
fix @schema -> nullable not take effect in property

### DIFF
--- a/oas-contract-tests/src/main/java/springfox/test/contract/oas/model/User.java
+++ b/oas-contract-tests/src/main/java/springfox/test/contract/oas/model/User.java
@@ -175,7 +175,7 @@ public class User {
    *
    * @return email
    */
-  @io.swagger.v3.oas.annotations.media.Schema(example = "john@email.com", description = "")
+  @io.swagger.v3.oas.annotations.media.Schema(example = "john@email.com", description = "", nullable = true)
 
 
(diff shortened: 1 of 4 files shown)


oas-contract-tests/src/main/java/springfox/test/contract/oas/model/User.java
```diff
@@ -175,7 +175,7 @@ public User email(String email) {
    *
    * @return email
    */
-  @io.swagger.v3.oas.annotations.media.Schema(example = "john@email.com", description = "")
+  @io.swagger.v3.oas.annotations.media.Schema(example = "john@email.com", description = "", nullable = true)
```
aimiebell, line 178: I detect that this code is problematic. According to the [Bad practice (BAD_PRACTICE)](https://spotbugs.readthedocs.io/en/stable/bugDescriptions.html#bad-practice-bad-practice), [Nm: Method names should start with a lower case letter (NM_METHOD_NAMING_CONVENTION)](https://spotbugs.readthedocs.io/en/stable/bugDescriptions.html#nm-method-names-should-start-with-a-lower-case-letter-nm-method-naming-convention).
Methods should be verbs, in mixed case with the first letter lowercase, with the first letter of each internal word capitalized.

Issue #3224, opened 2020-01-05
[Question - How to debug?] Swagger Recursively resolving dependencies for type X (Infinite loop)


terryBaz:
- [x] What version of the library are you using? latest version 2.9.2
- [x] What kind of issue is this? Question.

We wanted to use Swagger for an existing Spring-boot project ( > 70 model, > 250 endpoint) and we run into an issue where the application does not start and it enters in an infinite loop while setting up Swagger.

- Log
```
...
[DEBUG] 2020-01-04 23:28:32.500 Recursively resolving dependencies for type X;
[DEBUG] 2020-01-04 23:28:32.500 Adding type X2; for parameter x2
[DEBUG] 2020-01-04 23:28:32.501 Recursively resolving dependencies for type X2;
[DEBUG] 2020-01-04 23:28:32.501 Adding type Ljava/util/List<X3>; for parameter x3s
[DEBUG] 2020-01-04 23:28:32.501 Adding collectionElement type X3;
[DEBUG] 2020-01-04 23:28:32.501 Recursively resolving dependencies for collectionElement type X3;
[DEBUG] 2020-01-04 23:28:32.501 Adding type Ljava/util/List<X4;>; for parameter x4s
[DEBUG] 2020-01-04 23:28:32.501 Adding collectionElement type X4;
[DEBUG] 2020-01-04 23:28:32.501 Recursively resolving dependencies for collectionElement type X4;
....
```
- POM
```
<parent>
    <groupId>org.springframework.boot</groupId>
    <artifactId>spring-boot-starter-parent</artifactId>
    <version>2.1.3.RELEASE</version>
    <relativePath /> <!-- lookup parent from repository -->
</parent>
...
    <dependencies>
        <dependency>
            <groupId>io.springfox</groupId>
            <artifactId>springfox-swagger2</artifactId>
            <version>2.9.2</version>
        </dependency>
        ....
    <dependencies>
```
Any help appreciated, even some hints of how to dig into this problem, because the log is not helpful and I do not know where to start! Please let me know if you need more details.

Ps: I also dropped a question in [SO](https://stackoverflow.com/questions/59598383/swagger-recursively-resolving-dependencies-for-type-infinite-loop)

frecco75:
Hello @tarekbazine, I have a proposal solution for this issue (I was in a similar situation).
The PR is https://github.com/springfox/springfox/pull/3728

Could you try it with your use case and check if the problem is fixed?

tarekbazine:
[...]

PR #3728, opened 2021-02-10
fix infinite loop (memory leak) when recursive schemas

frecco75:
#### What's this PR do/fix?
This PR fixes recursive infinite loop (and memory leak) when there are a lot of schemas depending on each other.

Use multiple models where 
A depends on B and C
C depends on D
D depends on B 
etc ...

The problem was we have to distinguish the read phase of resolvedType into modelContexts and the processed phase.
We should not create many instances of ResolvedType for the same model if it is a dependency of multiple other models.

Work :
- `seen` method has been renamed to `processed`
-  `hasSeenBefore` method has been renamed to `isProcessed`
- new methods seen / hasSeenBefore

https://github.com/springfox/springfox/issues/3224

### DIFF
--- a/springfox-schema/src/main/java/springfox/documentation/schema/DefaultModelDependencyProvider.java
+++ b/springfox-schema/src/main/java/springfox/documentation/schema/DefaultModelDependencyProvider.java
@@ -77,7 +77,6 @@ public class DefaultModelDependencyProvider implements ModelDependencyProvider {
   @Override
   public Set<ResolvedType> dependentModels(ModelContext modelContext) {
     return Stream.concat(resolvedDependencies(modelContext).stream()
-            .filter(ignorableTypes(modelContext))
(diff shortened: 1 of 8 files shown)


kerbart (approved):

frecco75:
Is it possible to examine this fix ?

It's a well known problem as listed in the Q&A chapter of the documentation and this fix is a stable solution; we don't need anymore to use a workaround with `docket.ignoredParameterTypes` ignoring the type.

> **Q. Infinite loop when springfox tries to determine schema for objects with nested/complex constraints?**
>
> If you have recursively defined objects, I would try and see if providing an alternate type might work or perhaps even ignoring the offending classes e.g. order using the docket. ignoredParameterTypes(Order.class). This is usually found in Hibernate domain objects that have bidirectional dependencies on other objects.

PR #3736, opened 2021-02-18
Support relative server url

### LINKED ISSUE #3735: Support relative server url
## Overview

* Version: 3.0.0
* Type: Feature request

When requesting swagger api documentation, Springfox always return inferred server url (processed in `WebMvcBasePathAndHostnameTransformationFilter` or `WebFluxBasePathAndHostnameTransformationFilter`). I propose that Springfox can return relative server url.

## Use Case

In our micro-service platform, applications can not be accessed directly and must be routed by gateways (based on Nginx). One gateway always contains multiple applications and each application has a separate route prefix.

For example, a Springfox application with a /login api is routed by a gateway under `http://intranet.host/app1/`. Then the Swagger UI access url should be `http://intranet.host/app1/swagger-ui/` and /login api endpoint should be `http://intranet.host/app1/login`.

But currently the inferred server url is `http://intranet.host/` and the /login api endpoint is `http://intranet.host/login` (missing /app1/ route prefix). This is because Springfox infers server url from request url, and the request url it sees is the one gateway sends to application (`http://xxx.xxx.xxx.xxx/v3/api-docs`), instead of the one that browser sends to gateway (`http://intranet.host/app1/v3/api-docs`).

Sample Nginx config:

```nginx
location /app1/ {
  proxy_pass http://127.0.0.1:8080/;
  proxy_set_header Host $http_host;
  proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
  proxy_set_header X-Forwarded-Proto $scheme;
}
```

## Workaround

Currently I use the following workaround to solve the problem:

```java
@Configuration
@Order(Ordered.HIGHEST_PRECEDENCE + 5)
public class SwaggerServerUrlFilter implements WebMvcOpenApiTransformationFilter {
  @Override
  public OpenAPI transform(OpenApiTransformationContext<HttpServletRequest> context) {
    OpenAPI openApi = context.getSpecification();
    HttpServletRequest request = context.request().orElse(null);
    if (request != null && CollectionUtils.isNotEmpty(openApi.getServers())) {
      String referer = request.getHeader(HttpHeaders.REFERER);
      if (StringUtils.contains(referer, "/swagger-ui/")) {
        String baseUrl = StringUtils.substringBeforeLast(referer, "/swagger-ui/") + "/";
        for (Server server : openApi.getServers()) {
          server.setUrl(baseUrl);
        }
      }
    }
    return openApi;
  }
```

* Register a custom filter bean which has lower priority than `WebMvcBasePathAndHostnameTransformationFilter` (this means that it will be called after the latter one)
* Determine whether the request is sent from Swagger UI page by checking the referer HTTP header
    * If it is, replace the inferred server url with the one extracted from referrer

It works, though not elegant.

## Propose

To solve the problem, it's better to return relative server url and let swagger client (or browser) resolve it to absolute url. We might also guess the original request url from referer HTTP header (or force clients to send original request url as a parameter) and then build an absolute url in the server side, but that might be too complicated and not reliable enough.

Fortunately [OpenAPI specification](http://spec.openapis.org/oas/v3.0.3#server-object) allows relative server url:

> This URL supports Server Variables and MAY be relative, to indicate that the host location is relative to the location where the OpenAPI document is being served

Previously swagger-ui cannot process relative server url correctly, but this has been fixed in https://github.com/swagger-api/swagger-js/pull/1883/. We just need to bump the swagger-ui version to 3.43.0 or above.

Proposed changes:

* Bump the swagger-ui version to 3.43.0 or above
* Allow users to specify custom server url more easily. Currently users can only specify custom server url by providing a custom `WebMvcBasePathAndHostnameTransformationFilter` bean. This is a little tedious.
    * Option 1: Allow users to specify custom server by `Docket#servers` programmatically and Springfox should stop inferring server url when custom server url is given. Currently `Docket#servers` is not passed to `OpenAPI#servers` (`ApiDocumentationScanner#scan` doesn't fill `Documentation#servers`)
    * Option 2: Allow users to specify custom server by configuration option (say `springfox.documentation.server-url`)
    * Option 3: Add a boolean configuration option (say `springfox.documentation.relative-server-url`), and check it in `WebMvcBasePathAndHostnameTransformationFilter`

bianjp:
Resolve #3735

### DIFF
--- a/docs/asciidoc/current-documentation.adoc
+++ b/docs/asciidoc/current-documentation.adoc
@@ -97,6 +97,24 @@ allows the properties to be  overridden
 |2.0             | springfox.documentation.swagger.v2.path
 |=======================
 
+=== Customizing the server url.
+
+By default, Springfox infers server url from the request url of swagger endpoint (say /v3/api-docs).
+
+This may not work if your application is behind a proxy with path prefix. In this case, you can specify a custom server url when building docket:
+
+.Custom server url configuration
+[source,java,linenums]
+----
+  @Bean
+  public Docket swaggerSpringMvcPlugin() {
+    return new Docket(DocumentationType.OAS_30)
+            .servers(new ServerBuilder().name("relative").url("../").description("Relative server url").build());
+  }
+----
+
+NOTE: Swagger client will resolve relative server url against swagger endpoint (say /v3/api-docs) instead of the Swagger UI url (say /swagger-ui/) you are visiting in Browser.
+
 === Configuring startup.
 If you'd like to delay the startup of springfox, you could choose to set auto-startup to false. The property to use
 is `springfox.documentation.auto-startup` and this could either be passed in as a `-D` jvm arg or via a property in
--- a/springfox-oas/src/main/java/springfox/documentation/oas/web/WebFluxBasePathAndHostnameTransformationFilter.java
+++ b/springfox-oas/src/main/java/springfox/documentation/oas/web/WebFluxBasePathAndHostnameTransformationFilter.java
@@ -27,6 +27,9 @@ public class WebFluxBasePathAndHostnameTransformationFilter implements WebFluxOp
   @Override
   public OpenAPI transform(OpenApiTransformationContext<ServerHttpRequest> context) {
     OpenAPI openApi = context.getSpecification();
+    if (openApi.getServers() != null && !openApi.getServers().isEmpty()) {
+      return openApi;
+    }
     context.request().ifPresent(request -> {
       String requestUrl = decode(new ForwardedHeaderTransformer().apply(request).getURI().toString());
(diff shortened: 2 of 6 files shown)


luowenqiang:
This problem may be caused by the design problem of ForwardedHeaderExtractingRequest class. In the adjustedRequestURL method, the requestUrl variable parsed by forwardedPrefixExtractor is not used.
Here's my solution:
```java
public String adjustedRequestURL() {
    if (StringUtils.isBlank(this.forwardedPrefixExtractor.getRequestUri())) {
        return String.format("%s://%s:%s",
                getScheme(),
                getServerName(),
                getServerPort());
    }

    return String.format("%s://%s:%s%s",
            getScheme(),
            getServerName(),
            getServerPort(),
            getRequestURI());
}
```

esotericman:
I agree @luowenqiang ,you could add a context by condition @bianjp

PR #3786, opened 2021-03-31
fix noopValidator loggerName

### LINKED ISSUE #3785: NoopValidator class always logs debug even if springfox logger is lower level
This line seems to prevent logger from being configured whit all springfox other classes
https://github.com/springfox/springfox/blob/ab5868471cdbaf54dac01af12933fe0437cf2b01/springfox-core/src/main/java/springfox/documentation/builders/NoopValidator.java#L13

so you will always have debug logs for this class, because logger name it is not prefixed with springfox or package name

should be 
`private static final Logger LOGGER = getLogger(NoopValidator.class);`
as it was before

kyuuk:
this was preventing NoopValidator logger from being configured with all springfox classes


kyuuk:
up

Issue #3503, opened 2020-08-17
@ApiResponse response parameter not used in 3.0.0


varjasz:
Starting with 3.0.0 (in 2.9.2 it worked properly)

`@ApiResponse(code = 404, message = "Not Found", response = CustomError::class)`

The 404 - Not Found is in the documentation, but the defined response (CustomError in example) not shown in the generated documentation (response sample and model)

dilipkrish:
Thanks for reporting @varjasz

gonwan:
Same issue here.
After some debugging, root cause should be: https://github.com/springfox/springfox/blob/07a4c86d70b791f21c2d2f656a863920b8c883c7/springfox-swagger2/src/main/java/springfox/documentation/swagger2/mappers/ServiceModelToSwagger2Mapper.java#L117

Springfox 3.0 uses v3 models by default, but `source.getResponses()` gives wrong type. To workaround it for now, add:
`springfox.documentation.swagger.use-model-v3=false` in your application.properties.

drognisep:
> To workaround it for now, add: springfox.documentation.swagger.use-model-v3=false

The prescribed workaround doesn't seem to work for `ResponseEntity<?>` return types. I'm not sure if that's a separate issue or if it's part and parcel to this bug.

DominicIovino:
I can confirm the fix with `springfox.documentation.swagger.use-model-v3=false` for doc type `DocumentationType.SWAGGER_2`.

On the other hand, shouldn't it work out of the box with doc type `DocumentationType.OAS_30`? I tried it both with and without the app.prop setting but it behaves the same as the initial issue by not showing up at all.

arlyon:
Also getting this on springfox 3.0.0 using OAS_30. Specifically, the problem is that my POST requests to a JPA respository return a body with 201 responses, which are incorrectly labelled to have no body. Overriding this using `ApiResponse` has no effect.

**edit**: I believe the issue for me comes from here: 

https://github.com/springfox/springfox/blob/ab5868471cdbaf54dac01af12933fe0437cf2b01/springfox-swagger-common/src/main/java/springfox/documentation/swagger/readers/operation/SwaggerResponseMessageReader.java#L187-L198

[...]

PR #3867, opened 2021-07-23
Add json media type for EntitySaveExtractor

arlyon:
#### What's this PR do/fix?

When using JPA and the `EntitySaveExtractor` to generate documentation, the `produces` field is not set causing `SwaggerResponseMessageReader` not to generate a proper response for this type. The side-effect of this is that `@ApiResponse` directives are ignored. Note: I am assuming from the [spring data rest docs](https://docs.spring.io/spring-data/rest/docs/current/reference/html/#repository-resources) that it is okay to hard-code the media type as `application/json`, but happy to be proven wrong

No unit tests, but can add some. To verify, put a custom `ApiResponse` directive on a JPARepository

```
public interface SkillRepository extends JpaRepository<Skill, Long> {
  @Override
  @ApiResponses({@ApiResponse(code = 201, message = "Created", response = EntityModelSkill.class)})
  <S extends Skill> S save(S user);

  class EntityModelSkill extends EntityModel<Skill> {

  }
}
```

Without these changes, the ApiResponse directive is ignored, whereas with them the response type is correctly declared in the resulting schema, on PUT, POST, and PATCH.

Thought process is documented on my comment on this issue: https://github.com/springfox/springfox/issues/3503#issuecomment-884230856

I don't believe this solves the issue outright, but does fix a critical piece of functionality when using JPA. Next up is to figure out why 201 responses on POST don't declare the return type by default, but declaring the `ApiResponse` with the class above manually is a good enough work-around in the meantime.

 https://github.com/springfox/springfox/issues/3503

### DIFF
--- a/springfox-data-rest/src/main/java/springfox/documentation/spring/data/rest/EntityFindOneExtractor.java
+++ b/springfox-data-rest/src/main/java/springfox/documentation/spring/data/rest/EntityFindOneExtractor.java
@@ -35,19 +35,15 @@ class EntityFindOneExtractor implements EntityOperationsExtractor {
 
     context.crudMethods().getFindOneMethod()
         .map(method -> new HandlerMethod(context.getRepositoryInstance(), method))
-        .ifPresent(handler -> {
-
-          entityAction(context, handler)
-              .path(String.format("%s%s/{id}",
-                                  context.basePath(),
-                                  context.resourcePath()))
-              .supportsMethod(GET)
-              .parameterType(ParameterType.ID)
-              .build()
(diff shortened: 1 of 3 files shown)


DDERGOU:
你好，邮件已收到，谢谢！

arlyon:
Still waiting for a merge, ty bot

arlyon:
Still waiting for a merge, ty bot

arlyon:
Still waiting for a merge, ty bot

arlyon:
Still waiting for a merge, ty bot

Issue #2389, opened 2018-04-30
Support for documentation from property file for all annotations/attributes


MichaelSokolich:
- Using Version 2.8.0

- Question: Reading the documentation, I see support for documentation from property file lookup is only supported for @ApiParam#value(), @ApiImplicitParam#value(), @ApiModelProperty#value(), @ApiOperation#value(), @ApiOperation#notes(), @RequestParam#defaultValue(), @RequestHeader#defaultValue(). Is there any plan to add support for other annotation attributes. For example, we have a properties file that stores all the response messages for different errors and would love to reference the properties file rather than hand-type and manually update if the message changes. Like @ApiResponse(code = 404, message = "{error.auth.user.notFoundByLogin}").

dilipkrish:
Thats a great idea. The initial work was thanks to @jfiala work to parameterize descriptions via property files. Would certainly welcome PRs for that.

MichaelSokolich:
I have a lot of work to get done with pressing deadlines that I cannot work on this now but this is on my company's nice-to-have's list. Once core issues are solved if this is still open I'll try and work on an implementation.

dilipkrish:
@MichaelSokolich thats great news! Thank you!

cseunghoon7:
Is this issue still open ? I also want this feature to support more annotation attribute such as @ApiResponse(code=404, message=""), @ApiModel(description=""). If nobody started to help yet, I'll try.
 
@dilipkrish please, let me know

PR #3884, opened 2021-08-12
Added property file replacement for several annotation parameters

Deroin:
#### What's this PR do/fix?

This extends the capability of getting texts from property files for certain annotations. In particular

- `@ApiResponse#message()`
- `@ApiModel#description()`
- `@ExampleProperty#value()`
- `@ExampleObject#value()`

I didn't know if there are any other annotations which may benefit from this. If you know any other feel free to tell me and I'll include this into this PR. 

I looked for unit tests that do similar things but couldn't find any, thus I didn't add any either. But I checked this out in a side project. To test it manually simply create a properties file with messages and use them for the above listed annotation parameters as described in http://springfox.github.io/springfox/docs/current/#property-file-lookup.

https://github.com/springfox/springfox/issues/2389

### DIFF
--- a/springfox-swagger-common/src/main/java/springfox/documentation/swagger/readers/operation/OpenApiOperationTagsReader.java
+++ b/springfox-swagger-common/src/main/java/springfox/documentation/swagger/readers/operation/OpenApiOperationTagsReader.java
@@ -18,16 +18,13 @@
  */
 package springfox.documentation.swagger.readers.operation;
 
+import static java.util.stream.Collectors.toCollection;
+import static java.util.stream.Collectors.toSet;
+import static springfox.documentation.service.Tags.emptyTags;
+
 import io.swagger.v3.oas.annotations.Operation;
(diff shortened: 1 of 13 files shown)


svrsaljko:
Thank you very much. @ApiResponse "message" from properties file now works for me!
From my side I can suggest:
-  @ Tag (io.swagger.v3.oas annotation @Api is deprecated, this is now for desciribing Api in swagger-ui) "name" and "description"  for properties file
-@ApiModelProperty "value" don't work for me, I don't know if this is error in documentation, documentation says that http://springfox.github.io/springfox/docs/current/#overriding-descriptions-via-properties 
"value" and "description" should work although "description" doesn't even exist as prop in source code, do you maybe have problem with that also? 

I opened issue about that.
https://github.com/springfox/springfox/issues/3895

Deroin:
Extended this to include `name` and `description` from `@io.swagger.v3.oas.annotations.tags.Tag`

PR #3935, opened 2021-11-22
Update Spring Boot dependency to 2.6.0

CynanX:
The newest version of Spring Boot 2.6.0 changes the way in which the default strategy for matching request paths against registered Spring MVC handler mappings. This has changed from AntPathMatcher to PathPatternParser which causes an exception to be thrown when using with SpringFox v3.0.0.

This change updates the SpringFox dependency of Spring Boot to v2.6.0 and the accompanying files required to support the new version.


cpesch:
There hasn't been a commit since Oct 2020. Is the project dead? The maintainer disappeared?

CynanX:
I'm now migrating over to SpringDocs instead.

PR #3936, opened 2021-11-29
Attempting to fix spring upgrade to 5.3.13

mbazos:
#### What's this PR do/fix?
Attempting to upgrade spring framework to the latest and spring boot

No I haven't done any yet but not sure if they are required as current tests might cover this change.

Latest springframework is deprecating `PatternsRequestCondition` and it's no longer the default https://github.com/spring-projects/spring-boot/wiki/Spring-Boot-2.6-Release-Notes#pathpattern-based-path-matching-strategy-for-spring-mvc

springfox swagger won't load when using the latest springframework or spring-boot

### DIFF
--- a/gradle/dependencies.gradle
+++ b/gradle/dependencies.gradle
@@ -19,13 +19,13 @@ ext {
   slf4j = "1.7.25"
   snakeyaml = '1.26'
   spock = "1.3-groovy-2.5"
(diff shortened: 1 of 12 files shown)


quaff:
Could you squash commits as one and force push?

gtiwari333:
Any idea when this will be merged?

